PR #3, opened 2015-11-27
Wrapped everything up in a package and implemented tests using pytest

nils-werner:
We've seen your idea was greeted by the NumPy folks with enthusiasm but unfortunately discussion quickly faded afterwards. Your project is quite helpful in speeding up our beta NTF experiments so we figured we might give something back.

This PR

 1. shoves things around a bit so you can make a package out of it.
    It makes installing and using it much easier as, if case you decide to upload it to PyPI, people can then simply install it using `pip install opt_einsam` and `import opt_einsum`.
 2. We have recently discovered pytest and figured it would help reduce the boilerplate needed in your `numpy_test.py` script.

To install and test it you can simply do

    $ pip install -e .[tests]
    $ py.test -v

I didn't have a close enought look at the `test_*` files to see how they could be implemented in a way that can be run and evaluated automatically so I simply moved them to `compare/`, so that pytest doesn't stumble across them :-)


dgasmith:
Thank you! This is a really great addition to the project and I am very happy that this was useful for others.

Pushing this repo out to PyPI as a module is probably a good move for now. I was pretty bogged down for these past six months, but hopefully I can get back to this project and refine a few areas. I checked out the Numpy thread and it appears I missed some movement over there. I might try to get this merged again.

Most the `test_*` files are for debugging and comparison individual times so moving them to a compare folder is probably the right move.

Let me know if there are any features that you would like.

PR #5, opened 2016-03-23
Making opt_einsum python 3 compatible

faroit:
The code had a minor issue with python 3, mainly because of using `basestring` instead of `str`. This is fixed now so that the code runs on python 2.7 as well as python 3.5. 

I also did reformat some parts to make the code compatible to the PEP8 recommendations.


faroit:
I would also suggest to bump up the version after you reviewed and possibly merge the code, so that other  project (we use it here https://github.com/aliutkus/commonfate) can refer to the new python 3 compatible version.

dgasmith:
@faroit Everything looks great, thanks for fixing this up.

Glad you are finding this project helpful. Out of curiosity have you run into any bugs or needed features?

faroit:
For us it works perfectly as a drop-in replacement of ```einsum```

Issue #7, opened 2017-09-13
Optimal path with constant arguments


shoyer:
I have an application where there is a large tensor contraction in the inner loop, with some arrays that are fixed and others that are updated between iterations. Operations only involving the constant terms could done ahead of time (outside the loop), but this may or may not actually improve performance.

It could be nice if opt-einsum had a way to mark some arguments as constant, in which case the cost of contracting them them could neglected when computing the cost of paths only involving these arguments.

dgasmith:
@shoyer Interesting, can you write up a small example that I can hack on?

shoyer:
I don't know how *interesting* an example it is, but here's the one that turned up in our project:
```
sizes = {'a': 29, 'b': 29, 'c': 4, 'd': 4, 'f': 3, 'i': 3, 'j': 10, 'o': 3, 's': 50}
einsum_string = 'sabj,sfab,ifo,cdj,siac->sobd'
```
The first three terms are fixed, the last two are vary in each iteration.

dgasmith:
Sorry it took me a bit to get back to this. If you look at the full contraction we see:

```
  Complete contraction:  sabj,sfab,ifo,cdj,siac->sobd
         Naive scaling:  9
     Optimized scaling:  6
      Naive FLOP count:  1.453e+09
  Optimized FLOP count:  1.482e+07
   Theoretical speedup:  98.028
  Largest intermediate:  5.046e+05 elements
--------------------------------------------------------------------------------
scaling   BLAS                  current                                remaining
--------------------------------------------------------------------------------
   6      TDOT          siac,ifo->focas                sabj,sfab,cdj,focas->sobd
   6     False        focas,sfab->ocabs                     sabj,cdj,ocabs->sobd
   6     False        ocabs,sabj->ocjbs                          cdj,ocjbs->sobd
   6      TDOT          ocjbs,cdj->sobd                               sobd->sobd
```

The possible intermediates to form are:
 - `sabj,sfab,ifo->objia` (N^7 scaling, 13x cost)
 - `sabj,sfab->abjf` (N^7 scaling, 12x cost)
 - `sfab,ifo->sabio` (N^6 scaling, 0.9x cost)
 - `sabj,ifo->sabjifo` (N^9 scaling, ...)

So it doesn't look quite worth it for this particular example, I think I know how to build something that automatically detects this however.

shoyer:
> So it doesn't look quite worth it for this particular example

[...]

PR #11, opened 2017-12-04
contract_expression for reusable contractions

### LINKED ISSUE #10: reusable einsum expressions
Essentially, I'm using ``opt_einsum`` for some tensor network calculations, and a significant proportion of time is spent in ``contract_path``, even for relatively large sized tensors and pre-calculated paths.

Specifically, a huge number of contractions with the same indices and shapes are performed over and over again and I already cache these using the path from ``contract_path``. However, it seems even with the path supplied, a large amount of time is spent parsing the input, in ``can_blas``, ``parse_einsum_input`` etc.

My suggestion would be something like:

```python
shapes = [(2, 3), (3, 4), (4, 5)]
my_expr = einsum_expression("ab,bc,cd->ad", *shapes)

for _ in many_repeats:
    ...
    x, y, z = (rand(s) for s in shapes)
    out = my_expr(x, y, z)
    ...
```

I.e. it would only accept arrays of the same dimensions (and probably an ``out`` argument) and would otherwise skip all parsing, using a ``contraction_list`` stored within the function.

Anyway, I'm about to knock something up to test for myself, but thought it might be of more general interest.

jcmgray:
This is a first run at ``contract_expression`` - for building reusable tensor contraction functions based solely on array shapes, which could then for example be cached for efficiency:

```python
>>> expr = contract_expression("ab,bc->ac", (3, 4), (4, 5))
>>> expr
<opt_einsum.contract.ContractExpression at 0x7f68f44801d0>
>>> a, b = np.random.rand(3, 4), np.random.rand(4, 5)
>>> c = expr(a, b)
>>> np.allclose(c, a @ b)
True
```

Closes #10 and might be a useful starting point for #7 (e.g. storing the constant args in the ``ContractExpression``).

Further thoughts
----

1. Semantics / name? The alternate einsum convention is possible but gets pretty unreadable: i.e. ``contract_expression((3, 4), (0, 1), (4, 5), (0, 1))``...
2. More checks? currently these are minimal (I've just tested to make sure that some error is thrown for wrong number of arrays/ranks, i.e. no silent failures). Would defeat the point to have too many run upon every evaluation.
3. Caching seems like such a natural extension that it might be worth including.
4. A stored representation of the contraction might be nice for pretty printing.

I did a little bit of linting/refactoring just to help me add stuff - hope thats alright!

### DIFF
--- a/README.md
+++ b/README.md
@@ -131,6 +131,35 @@ True
 
 By contracting terms in the correct order we can see that this expression can be computed with N^4 scaling. Even with the overhead of finding the best order or 'path' and small dimensions, opt_einsum is roughly 900 times faster than pure einsum for this expression.
 
+
+## Reusing paths using ``contract_expression``
+
+If you expect to repeatedly use a particular contraction it can make things simpler and more efficient to not compute the path each time. Instead, supplying ``contract_expression`` with the contraction string and the shapes of the tensors generates a ``ContractExpression`` which can then be repeatedly called with any matching set of arrays. For example:
+
+```python
+>>> my_expr = oe.contract_expression("abc,cd,dbe->ea", (2, 3, 4), (4, 5), (5, 3, 6))
+>>> print(my_expr)
+<ContractExpression> for 'abc,cd,dbe->ea':
+  1.  'dbe,cd->bce' [GEMM]
+  2.  'bce,abc->ea' [GEMM]
+```
+
+Now we can call this expression with 3 arrays that match the original shapes without having to compute the path again:
+
+```python
+>>> x, y, z = (np.random.rand(*s) for s in [(2, 3, 4), (4, 5), (5, 3, 6)])
+>>> my_expr(x, y, z)
+array([[ 3.08331541,  4.13708916],
+       [ 2.92793729,  4.57945185],
+       [ 3.55679457,  5.56304115],
+       [ 2.6208398 ,  4.39024187],
+       [ 3.66736543,  5.41450334],
+       [ 3.67772272,  5.46727192]])
+```
+
+Note that few checks are performed when calling the expression, and while it will work for a set of arrays with the same ranks as the original shapes but differing sizes, it might no longer be optimal.
+
+
 ## More details on paths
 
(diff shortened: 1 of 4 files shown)


dgasmith:
For your comments:
1) It could be good to have the string version here. We could overload `__repr__` for this.
2) This is a hard one. We could also check for shapes; however, folks might want to pass in larger or smaller tensors. There are quite a few checks internally for pair contractions as well. Perhaps we could consider wrapping this call with a `try/except` and say something like "Internal error occurred during .... This function should only be called for repetitive expressions and contains few checks.... Internal error was %s".
3) Lets get this in first, then we can look at memoizing various parts. Suggested to also memoize the path functions.
4) Agreed, maybe have this as `__str__` so that it prints out something reasonable.

I usually use 120 spaces for formatting. I added a `YAPF` file to the base directory with the required settings.

The PR overall looks great, and I agree that it could be useful for #7 as well.

jcmgray:
Oh and I can run it through yapf too. Did you mean that there should be a config file in the repo? Or just that 120lw + defaults is enough.

dgasmith:
I added a `yapf` file to master, but the changes to 120lw looks good.

This is looking really good. My only thought is if we want to shorten `__repr__` a bit from something like:
`ContractExpression('ab,bc,cd', contraction_list=[((2, 1), {'c'}, 'cd,bc->bd', ['ab', 'bd'], 'GEMM'), ((1, 0), {'b'}, 'bd,ab->ad', ['ad'], 'GEMM')], einsum_kwargs={})`
to:
`ContractExpression('ab,bc,cd->ad')

Im open either way. Once your happy with that I say pull this in.

jcmgray:
Yeah I was going for ``eval(repr(expr)) == expr`` but I'm not sure how helpful that identity actually is. Will change to the more compact form. And maybe add a small section to the readme as well?

dgasmith:
Adding something to the README would be great. I have started moving over to readthedocs, but have not yet completed the transfer.

jcmgray:
Okay think that about covers it. 

I'll just add that this shaved some decent time off the DMRG algorithms in my library ``quimb``, which were already pretty fast in large part due to ``opt_einsum``!

PR #12, opened 2018-02-03
Improvements for large networks of tensors

jcmgray:
This PR contains a couple of improvements for efficiently contracting large networks of tensors, e.g. MPS, PEPS, MERA.

1. Allow ~1000 unique indices to be used. This is only for python3, which uses unicode strings by default. The only trick here really is that calls (e.g. pairwise contractions) to np.einsum must have their string mapped into [a-z,A-Z] with a bit of care taken about ordering.

2. Improve the performance significantly when finding the greedy path for many tensors. This is partly from refactoring ``find_contraction``, which was a hot spot, but mainly from ignoring any potential outer products (which are never advantageous to perform eagerly) while potential inner products are remaining.

For example, finding a path for 200 tensors with ~300 unique indices takes < ~1sec. Although one eventually might want to speed things along with a few manual hints, this is pretty useful functionality.
And from a few small tests, it doesn't seem like much overhead has been introduced for small and/or heavily connected graphs.

Comments obviously welcome!

### DIFF
--- a/opt_einsum/parser.py
+++ b/opt_einsum/parser.py
@@ -3,9 +3,59 @@ A functionally equivalent parser of the numpy.einsum input parser
 """
 
 import numpy as np
+import sys
 
-einsum_symbols = 'abcdefghijklmnopqrstuvwxyzABCDEFGHIJKLMNOPQRSTUVWXYZ'
+einsum_symbols_base = 'abcdefghijklmnopqrstuvwxyzABCDEFGHIJKLMNOPQRSTUVWXYZ'
+einsum_symbols = str(einsum_symbols_base)
+
+# boost the number of symbols using unicode if python3
+if sys.version_info[0] >= 3:
+    einsum_symbols += ''.join(map(chr, range(193, 688)))
+    einsum_symbols += ''.join(map(chr, range(913, 1367)))
+
 einsum_symbols_set = set(einsum_symbols)
+
+
+def is_valid_einsum_char(x):
+    """Check if the character ``x`` is valid for numpy einsum.
+    """
+    return (x in einsum_symbols_base) or (x in ',->.')
+
+
+def has_valid_einsum_chars_only(einsum_str):
+    """Check if ``einsum_str`` contains only valid characters for numpy einsum.
+    """
+    return all(map(is_valid_einsum_char, einsum_str))
+
+
+def convert_to_valid_einsum_chars(einsum_str):
+    """Convert the str ``einsum_str`` to contain only the alphabetic characters
+    valid for numpy einsum.
+    """
+    # partition into valid and invalid sets
+    valid, invalid = set(), set()
+    for x in einsum_str:
+        (valid if is_valid_einsum_char(x) else invalid).add(x)
+
+    # get replacements for invalid chars that are not already used
+    available = (x for x in einsum_symbols if x not in valid)
+
+    # map invalid to available and replace in the inputs
+    replacer = dict(zip(invalid, available))
+    return "".join(replacer.get(x, x) for x in einsum_str)
+
+
+def find_output_str(subscripts):
+    """Find the output string for the inputs ``susbcripts``.
+    """
+    tmp_subscripts = subscripts.replace(",", "")
+    output_subscript = ""
+    for s in sorted(set(tmp_subscripts)):
+        if s not in einsum_symbols_set:
+            raise ValueError("Character %s is not a valid symbol." % s)
+        if tmp_subscripts.count(s) == 1:
+            output_subscript += s
+    return output_subscript
 
 
 def parse_einsum_input(operands):
@@ -45,7 +95,7 @@ def parse_einsum_input(operands):
         for s in subscripts:
             if s in '.,->':
                 continue
-            if s not in einsum_symbols:
+            if s not in einsum_symbols_set:
                 raise ValueError("Character %s is not a valid symbol." % s)
 
     else:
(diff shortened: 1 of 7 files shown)


dgasmith:
This looks very interesting! I am out at a conference for another week and a half. I'll see if I can get a review out this weekend, but it might be the week after that.

My first comment is that while exploring contraction optimization I did find it occasionally helpful to do small outer products to reduce the overall cost. Was this completely eliminated or is the option still available in the `greedy` code.

jcmgray:
Ah yes the situation is slightly more subtle than I thought regarding outer products, there is a good discussion in this paper: https://arxiv.org/abs/1304.6112. The place where this applies is essentially boils down to the following situation. Consider outer product ``A`` and ``B``, both of which only share indices with the rest of network ``C``. The expression is then

```python
einsum("i,j,ijk", A, B, C)
```

and if ``k>i`` and ``k>j`` then ``(AB)C`` is optimal. One can imagine a tree-like network where this might appear a lot. Some thoughts in no particular order:

1. ``'greedy'`` currently (i.e. not with this PR), it seems, only catches this when ``i`` and ``j`` are both 1 (haven't looked into why this is?):
```python
>>> from numpy.random import randn
>>> import numpy as np
>>> import opt_einsum as oe
>>> ni, nj, nk = 2, 2, 2**10
>>> A, B, C = randn(ni), randn(nj), randn(ni, nj, nk)
>>> einsum_str = "i,j,ijk->k"
```
```
>>> print(oe.contract_path(einsum_str, A, B, C, path='optimal')[1])
  Complete contraction:  i,j,ijk->k
         Naive scaling:  3
     Optimized scaling:  3
      Naive FLOP count:  1.229e+04
  Optimized FLOP count:  8.196e+03
   Theoretical speedup:  1.499
  Largest intermediate:  1.024e+03 elements
--------------------------------------------------------------------------------
scaling   BLAS                  current                                remaining
--------------------------------------------------------------------------------
   2     False                  j,i->ij                                ijk,ij->k
   3      GEMM                ij,ijk->k                                     k->k
```
```
>>> print(oe.contract_path(einsum_str, A, B, C, path='greedy')[1])  # not-optimal
Complete contraction:  i,j,ijk->k
         Naive scaling:  3
     Optimized scaling:  3
      Naive FLOP count:  1.229e+04
  Optimized FLOP count:  1.229e+04
   Theoretical speedup:  1.000
  Largest intermediate:  2.048e+03 elements
--------------------------------------------------------------------------------
scaling   BLAS                  current                                remaining
--------------------------------------------------------------------------------
   3      GEMM                ijk,i->jk                                  j,jk->k
   2      GEMM                  jk,j->k                                     k->k
```
... which would be easy behaviour to replicate by checking for scalars

2. I think in the most extreme case, ``ni=nj=1``, the maximum gain for a single instance is only ~2x speedup, which is not drastic - especially if one is not requesting the 'optimal' path. Once BLAS is used the difference seems to be quite minor if the few tests I've done -- would be good to see any tricky examples if you've got them though!

3. Skipping outer products could be done *only* when the expression is very large - hoping that this situation only arises towards the end of the contraction.

dgasmith:
1) Hmm, thats odd. If thats not working 3) isn't very useful.
2) Outer product + BLAS is usually a GEMV which isn't impressively more efficient than `einsum` due to memory bandwidth limitations.
3) I like that in the greedy algorithm, once the number of tensors crosses 10(?) the outer products are dropped.

jcmgray:
So the reason is that the ``removed_size`` of a outer product is always 1 and thus less than an inner product (or equal if the dimension is 1), and the ``greedy`` code prioritizes this over any flop cost.

I think this is sensible, there might be lots of cheap outer products which are overall bad, and knowing which *are* useful would require looking forward to subsequent potential contractions, like in the ``optimal`` code. [As an aside, this might be a nice mid-point between ``greedy`` and ``optimal``, you check every sequence of *two* contractions, the first of which can be an outer, then choose the best].

I think this also means that (apart from the case of size 1 indices), there is no change in the paths ``greedy`` is finding - outer products will still be optimized when they are all that remains.

jcmgray:
I've just added another change that improves performance quite significantly. The gist is, after each iteration of greedy, only calculate the contraction costs for the new tensor created by the best pairwise contraction, and just modify the rest (which we know are still the same cost). 

combined with the other tweaks this takes the cost of working out the path for my test 200 tensor expression from ~1min to ~100ms.

opt_einsum/paths.py
```diff
@@ -84,6 +81,56 @@ def optimal(input_sets, output_set, idx_dict, memory_limit):
     return path
 
 
+def _parse_possible_contraction(positions, input_sets, output_set, idx_dict,
+                                memory_limit, path_cost, naive_cost):
```
dgasmith: Can you add a snippet about what this internal function does?

opt_einsum/paths.py
```diff
@@ -84,6 +81,56 @@ def optimal(input_sets, output_set, idx_dict, memory_limit):
...
+        return None
+
+    # Add contraction to possible choices
+    return [sort, positions, new_input_sets]
+
+
+def _update_other_results(results, best, input_sets):
+    """Update the positions and provisional input_sets of ``results`` based on
+    performing the contraction result ``best``.
+    """
```
dgasmith: Do we have any checks to make sure the number of indices passed to einsum does not exceed what einsum can handle?

opt_einsum/parser.py
```diff
@@ -3,11 +3,61 @@
 """
 
 import numpy as np
+import sys
+
+einsum_symbols_base = 'abcdefghijklmnopqrstuvwxyzABCDEFGHIJKLMNOPQRSTUVWXYZ'
```
dgasmith: Can we build this string like below? `"''.join(map(chr, range(65, 91)))`

opt_einsum/paths.py
jcmgray: So ``_np_einsum`` ensures that only good characters are used for each contraction call to einsum. If there were more than 52 supplied to an individual einsum call then that would be a problem but this would I guess require a 52+ dimensional tensor if all contractions are pairwise!

opt_einsum/paths.py
jcmgray: Yes, will do.

opt_einsum/parser.py
jcmgray: One could write this as (to get both lower- and upper-case)
```python
''.join(map(chr, range(97, 123))) + ''.join(map(chr, range(65, 91)))
```
but I'm tempted to leave the current string as its shorter and a bit more obvious?

dgasmith:
Thanks for this great PR! I think we should mint a new release soon with these changes in there. Before that release goes out I would like to work a bit more on moving the docs over to an actual set of ReadTheDocs.

jcmgray:
Thanks, yes a pip-able version of this would be great. I have some recent brief experience with readthedocs if that can be of help.

PR #15, opened 2018-03-04
Update docs and tweak greedy

jcmgray:
Updates to the docs:
------------

- Added reusing paths section
- Added autosummary references for the main functions (so they are clickable in the docs)
- fixed a few rst warnings.

I was wondering if it might be a good idea to strip down the README a bit and have it point to readthedocs so as not to duplicate info?

Greedy tweak
-----------

The tweak to greedy is to calculate ``removed_size`` as the ``size_tensors_removed - size_tensor_created``. This largely produces the same contractions but is occasionally better --- on the two tests I had to update ``greedy`` now produces the same scaling as ``optimal``.

As a quick example of when this works well, imagine this tensor contraction (as a diagram with larger indices along the top and bottom)::

    A=A=A=A=A=A=A=A=A
    | | | | | | | | |
    O-O-O-O-O-O-O-O-O
    | | | | | | | | |
    B=B=B=B=B=B=B=B=B

without setting a custom ``memory_limit``, the original greedy wants to contract the entire top row, and then entire bottom row before doing anything else as these are the biggest indices, but this produces a huge tensor and is also slow.

With the new tweak this is curtailed naturally earlier on, and without setting a low ``memory_limit`` that might slow the contraction in other ways.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -61,3 +61,5 @@ docs/_build/
 target/
 
 \.pytest_cache/
+
+docs/source/autosummary/
--- a/.gitignore
+++ b/.gitignore
@@ -59,3 +59,5 @@ docs/_build/
 
 # PyBuilder
 target/
+
+\.pytest_cache/
--- a/README.md
+++ b/README.md
@@ -195,7 +195,7 @@ terms = ['a', 'a'] contraction = (0, 1)
 
 The most optimal path can be found by searching through every possible way to contract the tensors together, this includes all combinations with the new intermediate tensors as well.
 While this algorithm scales like N! and can often become more costly to compute than the unoptimized contraction itself, it provides an excellent benchmark.
-The function that computes this path in opt_einsum is called _path_optimal and works by iteratively finding every possible combination of pairs to contract in the current list of tensors.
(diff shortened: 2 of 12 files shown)


dgasmith:
Sorry (again) for the long hiatus. The docs look good.

I think the `memory_limit` change works as well. Trying to think of an optimization that would break this, but have to figure one out :). Have you come up with any possible complications there?

jcmgray:
Like I say, I think it mostly finds the same path -- I'm relying a bit on all the tests so haven't thought super hard about when it might be worse, but in practice it certainly seems to work nicely.

What I can say about when it is *different* is that this is mostly when you get a high rank tensor which has a big index, but smaller than the product of the rest of the indices, e.g:

```python
shape1 = (2, 2, 2, 2, 10) 
shape2 = (2, 2, 2, 2, 10)
einsum_str = 'abcdX,efghX'
```

Contracting that size 10 index starts to look much more like an outer product rather than the inner products that are meant to be favored, so it's best to look elsewhere for a contraction.

dgasmith:
Your example makes sense I think and doing that outer product is not likely favorable and in many ways should look more like an outer product than a inner. However, this is intuition rather than fact.

Can you add a small comment note where you made this change so that if this becomes a problem in the future we have some way of tracking it.

PR #17, opened 2018-03-15
non-numpy (e.g. GPU) backend support

jcmgray:
This adds non-numpy backend support to ``opt_einsum`` in two flavors: 

### General duck-typed support

For arbitrary ndarray libraries, i.e. if the arrays supplied have got a ``.shape`` attribute and opt_einsum can find ``{backend}.tensordot`` etc then it should work, returning whatever type the library uses. So for example 

```python
contract("abc,bcd,def->aef", *dask_arrays, backend='dask')
```

 will return a ``dask.Array`` without any explicit logic. I've added tests for [dask](https://github.com/dask/dask/issues/732) and also [sparse](https://github.com/pydata/sparse/issues/31).

Edit: I should add that extra functionality like taking diagonals etc is still reliant on a ``einsum`` implementation. Not sure about broadcasting status etc.

### Special conversion support

To support GPU operations on numpy arrays, here, on first call the arrays are converted to the correct placeholder, an expression generated using these, and then this expression can be called with normal numpy arrays, returning numpy arrays also. E.g.:

```python
expr = contract_expression("abc,bcd,def->aef", *shapes)
expr(*numpy_arrays, backend='theano')
```

Explicit tested support here is for ``tensorflow``, ``theano`` and ``cupy`` - GPU libraries where it makes sense to convert to and from.

### Adding more backends

``contract.py`` doesn't have any explicit reference to the various backends, thats all contained in ``backends.py`` now. So it should be very easy to add new backends. E.g. if ``tblis`` were to add ``tensordot``, it would fall under the duck-typing category above and I think no changes would actually be needed at all.

Obviously any recommendations welcome! Getting the right backend function and all that is cached so I don't think this has added any more than a few 100ns overhead.

### DIFF
--- a/opt_einsum/backends.py
+++ b/opt_einsum/backends.py
@@ -1,0 +1,136 @@
+"""
+Performing array conversions and operations with various backends.
+"""
+
+import importlib
+
+
+# known non top-level imports
+_aliases = {'dask': 'dask.array', 'theano': 'theano.tensor'}
+
+
+def _import_func(func, backend):
+    """Try and import ``{backend}.{func}``, raise an error if library is
+    installed but can't find that particular function.
+    """
+    try:
+        lib = importlib.import_module(_aliases.get(backend, backend))
+        return getattr(lib, func)
+    except AttributeError:
+        raise AttributeError("{} doesn't seem to provide the function {}".format(backend, func))
+
+
+# manually cache functions as python2 doesn't support functools.lru_cache
+_cached_funcs = {}
+
+
+def get_func(func, backend='numpy'):
+    """Return ``{backend}.{func}``, e.g. ``numpy.einsum``, cache result.
+    """
+    try:
+        return _cached_funcs[func, backend]
+    except KeyError:
+        fn = _import_func(func, backend)
+        _cached_funcs[func, backend] = fn
+        return fn
+
+
+# mark libs with einsum, else try to use tensordot/tranpose as much as possible
+_has_einsum = {}
+
+
+def has_einsum(backend):
+    """Check if ``{backend}.einsum`` exists, cache result for performance.
+    """
+    try:
+        return _has_einsum[backend]
+    except KeyError:
+        try:
+            get_func('einsum', backend)
+            _has_einsum[backend] = True
+        except AttributeError:
+            _has_einsum[backend] = False
+
+
+# Tensorflow
+
+def convert_arrays_to_tensorflow(arrays):
+    """Convert numpy arrays to ``tensorflow.placeholder`` instances.
+    """
+    import tensorflow
+    return [tensorflow.placeholder(x.dtype, x.shape) for x in arrays]
+
+
+def build_tensorflow_expression(arrays, expr):
+    """Build a tensorflow function based on ``arrays`` and ``expr``.
+    """
+    import tensorflow
+    placeholders = convert_arrays_to_tensorflow(arrays)
+    graph = expr._normal_contract(placeholders, backend='tensorflow')
+
+    def tensorflow_contract(*arrays):
+        session = tensorflow.get_default_session()
+        return session.run(graph, feed_dict=dict(zip(placeholders, arrays)))
+
+    return tensorflow_contract
+
+
+# Theano
+
+def convert_arrays_to_theano(arrays):
+    """Convert numpy arrays to ``theano.tensor.TensorType`` instances.
+    """
+    import theano
+    return [theano.tensor.TensorType(dtype=x.dtype, broadcastable=[False] * len(x.shape))() for x in arrays]
+
+
+def build_theano_expression(arrays, expr):
+    """Build a theano function based on ``arrays`` and ``expr``.
+    """
+    import theano
+    in_vars = convert_arrays_to_theano(arrays)
+    out_var = expr._normal_contract(in_vars, backend='theano')
+    graph = theano.function(in_vars, out_var)
+
+    def theano_contract(*arrays):
+        return graph(*arrays)
+
+    return theano_contract
+
+
+# Cupy
+
+def convert_arrays_to_cupy(arrays):
+    """Convert numpy arrays to ``cupy.ndarray`` instances.
+    """
+    import cupy
+    return [cupy.asarray(x) for x in arrays]
+
+
+def build_cupy_expression(_, expr):
+    """Build a cupy function based on ``arrays`` and ``expr``.
+    """
+    import cupy
+
+    def cupy_contract(*arrays):
+        cupy_arrays = convert_arrays_to_cupy(arrays)
+        cupy_out = expr._normal_contract(cupy_arrays, backend='cupy')
+        return cupy.asnumpy(cupy_out)
+
+    return cupy_contract
+
+
+# Dispatch to correct expression backend
+
+SPECIAL_BACKENDS = {
+    'tensorflow': build_tensorflow_expression,
+    'theano': build_theano_expression,
+    'cupy': build_cupy_expression,
+}
+
+
+def build_expression(backend, arrays, expr):
+    """Build an expression, based on ``expr`` and initial arrays ``arrays``,
+    that evaluates using backend ``backend``.
+    """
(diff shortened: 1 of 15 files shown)


opt_einsum/contract.py
```diff
@@ -414,13 +436,21 @@ def _core_contract(operands, contraction_list, **einsum_kwargs):
...
             # Build a new view if needed
             if (tensor_result != results_index) or handle_out:
+
+                transpose = tuple(map(tensor_result.index, results_index))
+
+                try:
+                    new_view = new_view.transpose(transpose)
+                except AttributeError:
+                    # some libraries don't implement method version
+                    new_view = _transpose(new_view, axes=transpose, backend=backend)
```
dgasmith: Can we write this `try/expect` in `_transpose` as well to make it a bit more straight forward?

opt_einsum/backends.py
```diff
@@ -0,0 +1,136 @@
...
+        cupy_out = expr._normal_contract(cupy_arrays, backend='cupy')
+        return cupy.asnumpy(cupy_out)
+
+    return cupy_contract
+
+
+# Dispatch to correct expression backend
+
+SPECIAL_BACKENDS = {
+    'tensorflow': build_tensorflow_expression,
```
dgasmith: Would it make things simpler if NumPy was just another backend?

opt_einsum/contract.py
```diff
@@ -397,7 +419,7 @@ def _core_contract(operands, contraction_list, **einsum_kwargs):
         handle_out = specified_out and ((num + 1) == len(contraction_list))
 
         # Call tensordot
-        if blas:
+        if blas or not has_einsum:
```
dgasmith: Not all operations can be written as BLAS, what happens if a backend does not have einsum?

tests/test_backends.py
```diff
@@ -0,0 +1,142 @@
...
+    found_dask = False
+
+try:
+    import sparse
+    found_sparse = True
+except ImportError:
+    found_sparse = False
+
+
+tests = [
```
dgasmith: Can we actually run these test through the normal test set?

opt_einsum/contract.py
jcmgray: Yep, that's definitely cleaner.

opt_einsum/contract.py
jcmgray: Yeah this is a bit unsatisfactory, my suggestion is this, have ``can_blas`` return ``False`` for 'can't blas' and ``None`` for 'prefer einsum' (i.e. blas still possible), then this condition will look like ``if blas or (blas is None and not has_einsum)``. Then any times where a backend doesn't work will raise an ``AttributeError`` saying the backend doesn't have einsum, which is preferable.

tests/test_backends.py
jcmgray: Yes, they are all on travis (I fixed an error with ``dask`` earlier), apart from ``cupy`` which actually requires CUDA / a  GPU to run.

opt_einsum/backends.py
jcmgray: I'll add ``numpy`` explicitly at the top to the non-special backends, and might rename these ones at the bottom as ``convert_backends`` which is a bit more apt.

jcmgray:
I've update things so that backends without ``einsum`` can do as many types of contraction as possible but will still error more clearly if they need einsum - this needed a little bit of a refactor of ``can_blas``.

There's also a built version of the added docs [here](http://opt-einsum-jcmgray.readthedocs.io/en/latest/backends.html) if you want to check that out.

By the way, I think ``greedy`` now scales as N^2, (N^2 initial cost plus N steps of checking <N contractions), was thinking I might just update those docs in this PR as well?

dgasmith:
Nice, I think the docs look good. Would it be too much to give an example for each backend along with labels so that users can skip right to that part they may be interested in?

For refactoring `can_blas` it might be nice to screen based on the type of BLAS operation it is. Perhaps like the following:
 - `False` or 0 - No BLAS possible
 - 1, 2, 3 - Standard BLAS level
 - 4 - BLAS operations that cannot be written as standard level N BLAS such as ('ijkl,jlab->ikab')

Sure please do update the greedy docs.

dgasmith:
Another optimization to consider and may do later is `numexpr`. It would be quite good for typical einsum like reductions.

jcmgray:
Sorry the delay! I think the docs recommendation is a good one for ease of access- will do that soon. Some other thoughts:

The BLAS recommendation also sounds nice - but I might leave for another time since here I think the requirement here is only 'should-blas', 'can-blas' or 'no-blas'.

``numba`` has just added support for transposing arbitrary ndarrays, so that will be an option too (would have to write a ``tensordot`` function, but it should be a bit faster than ``numpy``.

In general - and relating to ``numexpr`` - if one could perform the remaining ``einsum`` operations like axis reductions but without einsum, then you'd have a pretty complete replacement for einsum for any tensor library, which would just need some base requirement functions. i.e. something like:

```python
if can_blas:
    ...
    _tensordot(x, y, ...)
    ...
    _transpose(...)
    ...
else:
    _axes_reduction(x, y, ...)
```
where _axes_reduction just needs (I'm guessing) ``diag``, ``sum`` etc.
Are axis reductions the only non-tensordot-able contractions?

dgasmith:
I think axis reductions and Hadamard products are the only non-tensordot-able products (perhaps by definition?). That was my thought as well, we can fall back to numexpr or numba in certain cases and let the other backend handle the heavy lifting. My only concern with this is moving data on and off GPU's is quite expensive and it would be preferable to just keep the data on the GPU's if possible. I would say lets make an issue of this and get this PR in as is before it gets too large.

For the BLAS recommendation can you do a halfway change as I'm not very comfortable with `True`, `False`, `None` logic chains.

jcmgray:
I think if ``_axes_reduction`` would take a ``backend`` argument then it could just get its functions in the same way as ``_tensordot`` and GPU arrays would stay on the GPU etc. But anyway, agree it is a problem for a separate 'einsum without einsum' issue.

And sure, I'll make some changes to ``can_blas`` to make the three cases a bit more explicit, it's definitely a bit arcane at the moment.

dgasmith:
Agreed all around. Once the `can_blas` is committed I will pull this in. Also, the other PR will make codecov go away.

jcmgray:
Great, I've updated it to append ``'/EINSUM'`` to the blas type, e.g. ``'ab,cd->abcd'`` yields ``'OUTER/EINSUM'``. Does that seem alright? Had to steal a bit of space for the blas column in the path pretty printing is the only drawback.

Another more concise option might be ``'!OUTER'``.

I'll just do the docs, or you can pull this in now and I'll put that in a separate PR.

mrocklin:
Pretty slick.  I'm glad to see people building algorithms around the growing family of ndarray-compatible data structures out there.

cc @njsmith FYI

dgasmith:
@mrocklin Absolutely, we have built a fairly large C++/Python hybrid project where the internal tensors are in a custom format, but once they are reduced a reasonable size everything is routed through a NumPy based driver. Found [here](https://github.com/psi4/psi4).

We promote these ideas at the [Molecular Science Software Institute](http://molssi.org) as well.

njsmith:
This is pretty awesome.

If numpy allowed objects to override `tensordot`, would that be enough to make this work without the explicit `backend=` arguments? Or are there other operations needed too? (And what are they?)

CC: @shoyer

jcmgray:
It would certainly allow 'pure' einsum contractions. i.e. those where each index appears only twice including the RHS. Axes reductions ("aa->a") and hadamard products ("ab,ab->ab") still currently require einsum, though it simple situations, e.g. above, they could obviously be replicated using simpler functions.

It would also be fairly simple to explicitly detect e.g. dask arrays without the ``backend`` argument, but this isn't implemented and wouldn't be as general a solution.

dgasmith:
The `c_einsum` function would need to be overrides as well, although most of the work usually does fall on `tensordot`.

Issue #13, opened 2018-02-21
Einsum broadcast regression


dgasmith:
It was pointed out on the main Numpy repository that the current python-based parsing tech misses a few broadcasting operations and tries to run them through BLAS routines which does not work.

See the issue [here](https://github.com/numpy/numpy/issues/10343) and PR to fix it [here](https://github.com/numpy/numpy/pull/10352). This patch needs to be pulled over to opt_einsum before the next release.

PR #21, opened 2018-04-01
Grab bag of changes

dgasmith:
I had a bit of time to work on `opt_einsum` and ended up meandering through the code. Beyond poor git practices, this PR accomplishes the following:
 - Adds versioneer
 - Gives a minor overhaul to the docs (thanks @jcmgray they look good!)
 - Fixes issue #13 

I think the #13 fix is actually a bit better then what is in NumPy. I might need to port these changes over to NumPy soon.


jcmgray:
All looks nice. Is it worth linking to the docs at the top of the readme? Maybe with a 1 or 2 sentence summary of the package just before the bullet points?

dgasmith:
Good point and done.

PR #27, opened 2018-06-18
Allow arbitrary indices + fix minor perf regression when using ellipses

jcmgray:
Going over numpy/numpy#11345 I saw this line which would have become slow when all the new symbols were introduced. On my machine this new behaviour brings the time to find a simple contraction path with ellipses back down from about 80us to 35us.

This is not really necessary for the numpy code as it only has 52 symbols (are there plans to change that?).

Also with this change, the number of symbols used now should have no impact on performance, just memory, I was wondering about further increasing the number to several thousand?

### DIFF
--- a/docs/source/api.rst
+++ b/docs/source/api.rst
@@ -11,3 +11,4 @@ Function Reference
     opt_einsum.contract.ContractExpression
     opt_einsum.paths.optimal
     opt_einsum.paths.greedy
+    opt_einsum.parser.get_symbol
--- a/docs/source/ex_large_expr_with_greedy.rst
+++ b/docs/source/ex_large_expr_with_greedy.rst
@@ -17,7 +17,8 @@ that looks like::
 
 The meaning of this is not that important other than its a large, useful
 contraction. For ``n=100`` it involves 200 different tensors and about 300
-unique indices.
+unique indices. With this many indices it can be useful to generate them with
+the function :func:`~opt_einsum.parser.get_symbol`.
 
 Let's set up the required einsum string:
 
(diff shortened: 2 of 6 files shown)


dgasmith:
Unsure if `einsum` will additional characters. Big PRs like the one currently in take awhile so it might be good to bring it up again after the current PR goes through.

This is a good change to prevent that extra overhead. If there are no more performance regressions I would say go ahead and increase the total number of valid characters.

jcmgray:
OK I've gone ahead and increased the number of symbols to about 7500 along with another small performance improvement. The memory footprint (essentially ``oe.parser.einsum_symbols_set``) is about half a megabyte now. 

I've also chosen a range of characters that actually rendered as something on my system, but there's an argument for just choosing a contiguous range like ``''.join(map(chr, range(193, 8000)))`` as well.

dgasmith:
In some ways, does it matter if we accept all unicode chars? The performance bottlenecks seem to have been overcome. No reason to limit folks if they are say just grabbing the first `n` chars from `chr`.

jcmgray:
I think you're right - I've changed it so _any_ characters can be used which also essentially means any number of indices can be used. This simplified the code a bit I think as well.

To generate indices, rather than ``einsum_symbols`` there is now the function ``oe.parser.symbol(i)``. For readability, and to avoid the characters ``",->"`` and ``" "``, it runs through the alphabet characters then starts at ``chr(192)``.

Should this be put in the top namespace (``oe.symbol``) and documented?

jcmgray:
Open to renaming suggestions for ``symbol`` as well.

dgasmith:
I like moving this to top namespace with some docs. Perhaps `oe.get_symbol`?

jcmgray:
Yep ``get_symbol`` sounds good. I'll add some docs in the greedy docs and the large expressions example.

jcmgray:
I've made those changes. Adding examples for ``get_symbol`` involved specifying the file encoding to utf-8 for python2 - don't think this has any repercussions.

dgasmith:
Thanks for the updates, looks good overall. Just to double check your previous test is still on the ~35us mark even with the expansion of unicode chars?

jcmgray:
Yes to the timing, even a little bit quicker now due to not checking if the indices are valid now.

PR #32, opened 2018-08-04
add pytorch, tensorflow-eager support, minor optimizations

jcmgray:
## Description

There are 3 related changes here:

* The main change is the addition of backend support for [pytorch](https://pytorch.org) and [tensorflow's eager execution mode](https://www.tensorflow.org/guide/eager). Pytorch doesn't have tensordot yet (https://github.com/pytorch/pytorch/issues/8988, https://github.com/pytorch/pytorch/pull/10025), and it's ``transpose`` and ``einsum`` functions also don't match numpy's api so there's wrappers for each of those. All the backends are also in their own submodules now - it should be pretty clear and simple to add any others.

* If two contracted dimensions are *both* size 1 this is now is not marked as a 'broadcast' index, meaning that einsum is not required (good for performance, and also if a backend doesn't supply it).

* A variety of small optimisations, mostly changing for loops w/ list append to comprehensions, and also using ``np.broadcast_to(np.nan, shape)`` for dummy arrays.

-----

Some general notes on my experience of the performance of all the different GPU backends (compilation aside, all can offer pretty drastic speedup over numpy):

**Compiling backends [``tensorflow``, ``theano``]:**

- Compilation is slow (theano even more so than tensorflow) for large contractions, we are talking on the order of seconds
- Once compiled, the expression can be slightly faster to run (maybe 10-15%) than the best of the eager backends below
- Because of these two things - the eager backends below are generally much more preferable unless an expression is going to be extensively reused

**Eager backends [``cupy``, ``pytorch``, ``tensorflow`` in eager mode]:**

- ``cupy`` and ``pytorch`` both seem to have similar performance with eager-``tensorflow`` lagging behind a bit (could very well be specific to the GPU/setup I have)
- ``pytorch`` sadly doesn't have complex dtype support [yet](https://github.com/pytorch/pytorch/issues/755) so is less versatile for quantum/physics applications
- ``tensorflow`` eager mode needs to be enabled before anything else takes place (with ``import tensorflow as tf; tf.enable_eager_execution()``) which is a slight inconvenience

Will all that taken into account, ``cupy`` currently seems to be the easiest, most versatile GPU backend with excellent performance.


opt_einsum/backends/__init__.py
```diff
@@ -0,0 +1,17 @@
+from .dispatch import (get_func, has_einsum, build_expression, evaluate_constants, CONVERT_BACKENDS)
```
dgasmith: Missing description strings at the top of several of these files.

opt_einsum/contract.py
```diff
@@ -128,7 +128,7 @@ def contract_path(*operands, **kwargs):
 
     # Get length of each unique dimension and ensure all dimensions are correct
     dimension_dict = {}
-    broadcast_indices = [[] for x in range(len(input_list))]
+    bcast = set()
```
dgasmith: In this approach I believe we always keep a given index as broadcast even if it is "expanded" in a given contraction. This could lead to performance regression.

opt_einsum/contract.py
```diff
@@ -674,13 +663,11 @@ def __str__(self):
...
 
-class _ShapeOnly(np.ndarray):
+def shape_only(shape):
     """Dummy ``numpy.ndarray`` which has a shape only - for generating
     contract expressions.
     """
-
-    def __init__(self, shape):
-        self.shape = shape
+    return np.broadcast_to(np.nan, shape)
```
dgasmith: Interesting optimization by NumPy. Simpler to use this, nice change.

opt_einsum/tests/test_backends.py
```diff
@@ -36,6 +38,12 @@
 except ImportError:
     found_sparse = False
 
+try:
+    import torch
+    found_torch = True
+except ImportError:
+    found_torch = False
+
```
dgasmith: Could try using `pytest.importorskip` or similar through here. Might make things a bit simpler.

opt_einsum/contract.py
jcmgray: Yes, subclassing ``ndarray`` I think was having some side-effects when instantiated, whereas this is guaranteed to essentially allocate no memory

opt_einsum/contract.py
jcmgray: I don't think I quite understand this, do you have an example? I *think* what I did was just a refactor, apart from the case when an index is dimension 1 everywhere it appears (and is thus not broadcast), however there are many different possibilities with ``einsum`` so could have missed something!

opt_einsum/tests/test_backends.py
jcmgray: Yep that simplified things!

dgasmith:
Looks like torch is getting a segfault in travis. The libraries that we pull down are getting heavy enough that we should consider `conda`. If you do not have objections I can make a PR that switches to using `conda` over `pip`.

jcmgray:
No objections at all, ``conda`` certainly makes dependency management a lot easier. In a couple of other projects I install minconda to ``$HOME/conda`` which travis can then cache, speeds up installation a lot.

opt_einsum/contract.py
jcmgray: OK I think I see - essentially, we need to identify if an initially 'broadcasted' index becomes 'normal' at some point? I'll take a look at this and your solution in a bit!

jcmgray: So nearly all the logic can be handled by just relying on ``can_blas`` already (since it was modified a few PRs ago), including the above example. In fact, if I completely remove ``bcast`` logic from ``contract_path`` then all but 1 tests still pass! The missing case (``test_singleton_dimension_broadcast``) is essentially this:
```python
np.einsum("ij,jk->ik", p, p)  # p.shape = (1, 5)
``` 
which is something I wasn't aware ``numpy.einsum`` could do. My suggestion is therefore just to give ``can_blas`` a bit of extra information to catch this case, probably the dimensions involved.

dgasmith: Yes, that particular use case surprised me as well. The multitude of `np.einsum` edge cases is quite extensive.

Now that we have const expressions involved it likely does make more sense to move the logic to the `can_blas` section. Simply check that none of the shapes is 1 in the `removed_indices`.

I can add this later, but best to add the above example in tests by pulling out the einsum run data and checking if can_blas is True.

jcmgray: OK I've updated it so that the logic is in``can_blas`` which also allows 'healed' broadcast indices to blas. 

Just N.B doing this made me realize that the path algorithms assume broadcast indices are their full size I think, which is not necessarily true.

jcmgray:
Hmm yes it's working fine locally, and it was working with the pip installed torch earlier. There's a few things that come to mind that I'll try.

jcmgray:
The thing that seemed to be causing the segfault was using ``pytest.importorskip`` which is a bit weird. I've reverted back to the old test skip method for now.

jcmgray:
I've added one more tweak, which is to use the natural ``tensordot`` output order of indices (input order: ``"zx,xy->zy"``) rather than ``einsum`` output order (alphabetic: ``"zx,xy->yz"``) for **intermediaries** only - doesn't effect output. 

The rationale is just to minimise transpositions. This doesn't really matter for numpy as transposing is lazy, but it will be helpful for some other libraries (e.g. https://github.com/springer13/tcl) since it maintains the array contiguity throughout.

dgasmith:
Cool, looks good. I had thought that at some point we should make a second pass that goes through and minimizes transposes. This will help, but I think you can become a bit more optimal by forward "planning" of the output order.

I think this overall looks ready to go, is this true?

jcmgray:
> I think this overall looks ready to go, is this true?

Yes I think so.

PR #33, opened 2018-08-06
CI: Switches to conda for environment testing

dgasmith:
## Description
Switches testing to `conda` as the backend install are increasingly more compilation dependent.

## Questions
@jcmgray Do you have a conda caching example, historically this has given me quite a bit of trouble.

## Status
- [x] Ready to go

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -11,31 +11,33 @@ matrix:
     - python: 3.4
     - python: 3.5
     - python: 3.6
+    - python: 3.7
+
+  allow_failures:
+    - python: 3.7
 
 before_install:
+  # Additional info about the build
   - uname -a
   - free -m
   - df -h
   - ulimit -a
   - python -V
-  - pip install --upgrade pip setuptools
-  - pip install pytest pytest-cov codecov
-  - pip install numpy
-  - pip install codecov
-  - pip install tensorflow
-  - pip install theano
-  - pip install dask[array]
-  - pip install sparse
-  - pip install -e .
-  - if [ `python -c "import numpy; print(numpy.version.version)"` == "1.14.0" ]; then
-        pip install numpy --upgrade;
-    fi
+
+  # Setup python environment
+  - source devtools/travis-ci/before_install.sh
+  - python -V
+
+install:
+    # Install from environment
+  - conda env create -n test python=$PYTHON_VER -f devtools/conda-envs/full_test.yaml
+  - source activate test
+
+    # Build and install package
+  - python setup.py develop --no-deps
 
 script:
-  - py.test -v --cov=./
-
-# after_success:
-#   - ./tools/travis-upload-wheel.sh
+  - py.test -v --cov=opt_einsum opt_einsum/
 
 notifications:
     email: false
--- a/devtools/conda-envs/full_test.yaml
+++ b/devtools/conda-envs/full_test.yaml
@@ -10,6 +10,7 @@ dependencies:
   - tensorflow
   - dask
   - sparse
+  - theano
 
     # Testing
   - pytest
--- a/devtools/conda-envs/full_test.yaml
+++ b/devtools/conda-envs/full_test.yaml
@@ -1,0 +1,16 @@
+name: test
+channels:
+  - conda-forge
+dependencies:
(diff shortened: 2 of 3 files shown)


loriab:
Is the md5 stuff just to prove the integrity of the miniconda installer? Or for cache purposes ?

jcmgray:
> Is the md5 stuff just to prove the integrity of the miniconda installer? Or for cache purposes ?

Think it's just to verify the installer.

----

Regarding caching @dgasmith, for my ``xyzpy`` project I have a pretty simple [``.travis.yml``](https://github.com/jcmgray/xyzpy/blob/develop/.travis.yml) and [``travis-install-conda.sh``](https://github.com/jcmgray/xyzpy/blob/develop/deps/travis-install-conda.sh).

For the travis file you just add:
```yaml
cache:
  directories:
    - $HOME/conda
```
(or ``$HOME/miniconda`` etc.). And in the build script you essentially just check if ``$HOME/conda/bin`` exists - if it doesn't, install as usual - if it does, just update the packages.

These could be copied over quite directly - I think the only difference is the lack of md5 checking, and the support for the python matrix (should be quite easy to add, just remove python from the conda file and put it in the script with ``python=$PYTHON_VER``, I might add this to ``xyzpy`` now in fact!).

dgasmith:
I believe the hash is just for verification @lnaden can weight in a bit here. Levi, also check out the caching if you have a moment. It might work for the general case.

Not sure how landscape slipped into tests, nuked them from my auth keys.

Lnaden:
The MD5 Hash check is just for the Miniconda installer and its existence is for some very specific conditions:
* Ensures that someone has not hijacked the download link and somehow let you snag a miniconda which may report your Travis Secure envs (has not happened, but the check is basically a free action, so I've always just left it in. A bunch of copy-paste between projects, later it has just propagated)
* Prevent getting the wrong Miniconda version. Sometimes the Miniconda folks will update their download links for `latest` *before* updating the MD5 hashes, this can lead to some very unexpected behavior, especially on major Conda version upgrades, so this also traps that. (This has happened more times than I thought it should)
* Some of the projects I have worked on pinned to a maximal version of Conda, so we used that to make sure we did not upgrade due to bad links. Much less common now.

The caching is an interesting idea, but I don't think caching the conda install directory is worth it. As I understand how the Travis Caches work: Download directory, {do Travis things}, If directory has changed at all: re-cache.

Under that model, the directory would almost always re-cache again because of how frequently conda packages update. It may save some boot up time, but will almost always result in pulling down and re-caching, which can chew up time at the end if you are pushing the maximum.

Another reason you may not want to do that is testing philosophy. In the models I follow, you want your tests to replicate the use case of the most basic user of your code as possible. If that user meets the conditions of "Person who knows bare minimal of what your code does and the bare minimal of the supporting software ecosystem, and just wants to pick it up and use it," then you want to replicate the conditions of a user pulling down fresh the bare minimum to get up and going. That is just one style though and if practical Travis limits are becoming an issue, then maybe the Cache is the right option.

Just my opinion though.

Oh, and don't use the Caching system if you are using the Travis Xcode 7.3 image, its bugged in that version and they will not fix it due to it be deprecated.

jcmgray:
Those are some very good points regarding testing/caching - it sounds like it's probably best to leave it off for ``opt_einsum``.

I did originally use due it due to travis time constraints on a project with heavy build requirements, my experience is that it saves about 10mins per build - useful but nothing essential.

dgasmith:
Thanks @Lnaden for the additional descriptions, do you think we could merge parts of the discussion into the compchem cookiecutter?

On the time issue, it largely depends on how you deploy the conda setup. Running a full conda-build takes a very long time (15-20 minutes), but using builds from environment.yaml's and `python setup.py --deploy` has these tests running in 5 minutes flat.

Can @jcmgray or @loriab merge when comfortable with the changes?

Lnaden:
> do you think we could merge parts of the discussion into the compchem cookiecutter?

As in into the parent README.md and some comment strings? Sure!

dgasmith:
@Lnaden Yea, exactly where I was thinking.

PR #36, opened 2018-08-13
'auto' path mode + test coverage

jcmgray:
## Description
* Enables 'auto' mode as the default contraction path. This uses the optimal path for 4 or less operators, but otherwise the greedy path - @dgasmith not sure if you were also about to add 'auto' mode (in which case I can remove).

* A few other tweaks that up the test coverage a bit.

## Status
- [x] Ready to go

### DIFF
--- a/opt_einsum/helpers.py
+++ b/opt_einsum/helpers.py
@@ -117,7 +117,8 @@ def find_contraction(positions, input_sets, output_set):
     """
 
     remaining = list(input_sets)
-    idx_contract = set.union(*map(remaining.pop, sorted(positions, reverse=True)))
+    inputs = (remaining.pop(i) for i in sorted(positions, reverse=True))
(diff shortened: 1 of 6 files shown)


opt_einsum/helpers.py
```diff
@@ -117,11 +117,7 @@ def find_contraction(positions, input_sets, output_set):
     """
 
     remaining = list(input_sets)
-    idx_contract = set()
-
-    for i in sorted(positions, reverse=True):
-        idx_contract |= remaining.pop(i)
-
+    idx_contract = set.union(*map(remaining.pop, sorted(positions, reverse=True)))
```
dgasmith: Do we get much of a performance increase by rewriting this term as shown? Most of them are ok, but this one takes a bit to understand.

PR #41, opened 2018-08-17
Rename symbols to support torch.einsum

fritzo:
## Description
This renames symbols to support PyTorch 0.4.1 which supports only symbols a-z. This is useful when the overall computation has more than 26 dimensions but any single contraction requires only 26 or fewer dimensions.

## Questions
- [x] Do other backends work? I have only tested PyTorch. If other backends don't work, I'll simply limit the test to the PyTorch backend.

## Status
- [x] Ready to merge as soon as tests pass

## Tested

- Added a test with non-lowercase symbols
- Tested that this works in an application using Pyro https://github.com/uber/pyro/pull/1313

### DIFF
--- a/opt_einsum/backends/torch.py
+++ b/opt_einsum/backends/torch.py
@@ -5,7 +5,7 @@ Required functions for optimized contractions of numpy arrays using pytorch.
 from __future__ import absolute_import
 import numpy as np
 
-from ..parser import einsum_symbols_base
+from ..parser import einsum_symbols_base, get_symbol
 
 
 _TORCH_DEVICE = None
@@ -31,6 +31,12 @@ def transpose(a, axes):
 def einsum(equation, *operands):
     """Variadic version of torch.einsum to match numpy api.
     """
+    # rename symbols to support PyTorch 0.4.1 and earlier,
+    # which allow only symbols a-z.
+    symbols = sorted(set(equation) - set(',->'))
+    rename = {s: get_symbol(i) for i, s in enumerate(symbols)}
+    equation = ''.join(rename.get(s, s) for s in equation)
+
     torch, _ = _get_torch_and_device()
     return torch.einsum(equation, operands)
 
@@ -39,8 +45,6 @@ def tensordot(x, y, axes=2):
     """Simple translation of tensordot syntax to einsum.
     """
     # XXX: tensordot should be directly implemented in torch soon
-    torch, _ = _get_torch_and_device()
-
     xnd = x.ndimension()
     ynd = y.ndimension()
 
@@ -80,7 +84,7 @@ def tensordot(x, y, axes=2):
 
     # form full string and contract!
     einsum_str = "{},{}->{}".format(*map("".join, (x_ix, y_ix, out_ix)))
-    return torch.einsum(einsum_str, (x, y))
+    return einsum(einsum_str, x, y)
 
 
 def to_torch(array):
--- a/opt_einsum/backends/torch.py
+++ b/opt_einsum/backends/torch.py
@@ -5,8 +5,7 @@ Required functions for optimized contractions of numpy arrays using pytorch.
 from __future__ import absolute_import
 import numpy as np
 
-from ..parser import einsum_symbols_base, get_symbol
-
+from ..parser import convert_to_valid_einsum_chars, einsum_symbols_base
 
 _TORCH_DEVICE = None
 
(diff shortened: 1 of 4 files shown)


opt_einsum/backends/torch.py
```diff
@@ -31,6 +31,12 @@ def transpose(a, axes):
 def einsum(equation, *operands):
     """Variadic version of torch.einsum to match numpy api.
     """
+    # rename symbols to support PyTorch 0.4.1 and earlier,
+    # which allow only symbols a-z.
+    symbols = sorted(set(equation) - set(',->'))
```
dgasmith: Extreme, but can we throw if `len(symbols) > 26`.

opt_einsum/backends/torch.py
```diff
@@ -31,6 +31,12 @@ def transpose(a, axes):
...
+    rename = {s: get_symbol(i) for i, s in enumerate(symbols)}
+    equation = ''.join(rename.get(s, s) for s in equation)
```
dgasmith: @jcmgray We do this for `einsum` as well I believe. Is it time to have a single expression which converts from global to local symbols?

opt_einsum/backends/torch.py
```diff
@@ -39,8 +45,6 @@ def tensordot(x, y, axes=2):
     """Simple translation of tensordot syntax to einsum.
     """
     # XXX: tensordot should be directly implemented in torch soon
-    torch, _ = _get_torch_and_device()
```
dgasmith: @jcmgray Can you look over this as well?

fritzo: (Note the only reason I've removed this line is so I can reuse the `einsum()` wrapper in this file, rather than calling `torch.einsum` in two different places)

opt_einsum/backends/torch.py
fritzo: The advantage of not throwing is that this will continue to work when PyTorch fixes their einsum to allow more symbols. I'd rather let their current implementation throw.

dgasmith: Makes sense. I am not intimately familiar with the new `torch` symbols set, but would `opt_einsum.parser.convert_to_valid_einsum_chars` work here?

jcmgray:
> Do other backends work? I have only tested PyTorch. If other backends don't work, I'll simply limit the test to the PyTorch backend.

``einsum`` in numpy, cupy, dask and tensorflow (from only recently) all support upper case letters so tests should be fine. The ideal situation would be for pytorch to add support for upper case, but the implementation is c++ so its not super obvious to me how easy that would be (@t-vi?). Probably easier to add it here anyway!

---

It might be cleaner, and more useful for potential future backends, to modify the existing [machinery](https://github.com/dgasmith/opt_einsum/blob/7ef12a570ce2ba4b02a08cbe94e8cae02dd343be/opt_einsum/parser.py#L13): 
* ``opt_einsum.parser.convert_to_valid_einsum_chars``
* ``opt_einsum.parser.has_valid_einsum_chars_only`` 
* ``opt_einsum.parser.is_valid_einsum_char``

probably with a ``allow_uppercase`` keyword, which could just switch ``is_valid_einsum_char`` between checking against:
```python
import string
string.ascii_lowercase  # or
string.ascii_letters
```

Also, for what its worth, your strategy of just replacing all characters might be faster than replacing just the invalid ones, in which case do update ``convert_to_valid_einsum_chars`` with your snippet.

t-vi:
PyTorch's einsum makes some use of the fact that valid letters are consecutive and from a small fixed set (based on review comments when I previously dynamically allocated indices). It's not hard to extend the set and deal with several ranges but my impression is that lowercase a-z is all anyone ever uses in the wild.
That said, I still plan on having an extended greedy optimisation in ATen c++, maybe in a month or so.

fritzo:
> lowercase a-z is all anyone ever uses in the wild

We're using `opt_einsum.contract()` for hundreds of variables in Pyro. Each local contraction requires only a few variables (hence this PR), but we have one variable per time step in a Hidden Markov Model.

t-vi:
If there is a use case, supporting A-Z is easy. One could also consider a "post-parsing interface" (i.e. taking preprocessed equations) if that is useful. I must admit I always thought of einsum as a  convenient as hoc interface, so I learn something new here (thanks!).

dgasmith:
Closed/opened to trigger Travis. Once that passes and the `opt_einsum.parser.convert_to_valid_einsum_chars` is either used or updated this is ready to go.

dgasmith:
@fritzo Thanks for the PR, everything looks great and we will get this into the next release.

PR #43, opened 2018-08-20
Implement shared_intermediates context manager

### LINKED ISSUE #9: Re-use intermediates across various contractions (Common Subexpression Elimination)
Suppose you want to compute two contractions of the same tensors, e.g. contraction strings

`['ab,dca,eb,cde', 'ab,cda,eb,cde']`

The (globally) optimal way to do this would be to first perform the contractions over indices a,b and e, and then perform the remaining contractions over c and d for the two sets of contractions. The current opt_einsum implementation does not allow for such a _global_ optimization of contraction order and re-use of common intermediates.

I'm still organizing my thoughts on this, and all input would be most welcome. On a side note, I believe implementing such a more general optimization strategy will also fix #7 as a by-product.

### Optimization logic

A relevant publication suggesting a concrete algorithm for this optimization problem is

[_Hartono et al_, Identifying Cost-Effective Common Subexpressions to Reduce Operation Count in Tensor Contraction Evaluations](http://www.csc.lsu.edu/~gb/TCE/Publications/OpMinCSE-ICCS06.pdf)

I do not know to what extent the current code can be re-used in this more general setup, but the _single-term_ optimization should be under control with the current implementation.

### Interface

Such a multi_contract function could be called with a list of tuples (contractionString, [tensors, ..]), and would return a list of results with the same length as the input list.

Internally, one would have to find out which tensors are actually identical between the various contractions, and then use the above contraction logic. Ideally this information should be deduced automatically and not rely on user input being in the correct form. In the same spirit, dummy indices should be allowed to have arbitrary names, i.e. they should not have to match across different contractions to be correctly identified as a common subexpression.
This may require transforming the input into a 'canonical' form first to make sure that common subexpressions are captured correctly.

In contrast to the setup outlined in _Hartono et al_, contraction strings should maybe remain in their current 'simple' form and **not** be generalized to allow for numerical expressions like sums of tensors etc. Such a behavior can be implemented _a posteriori_ with the interface described here by computing the sum of the resulting tensors, e.g.

`contract('ab,ab + ab,ba', N,M) -->  'sum'(multi_contract( [('ab,ab', N,M), ('ab,ba', N,M)] ))`

Thus, restricting contraction strings to be of the form currently used does not cause loss of generality, the only downside being that it might lead to a slightly increased memory-footprint as the function would return several arrays instead of one.

Other thoughts?

fritzo:
Resolves #9 

## Description

This implements a `shared_intermediates` context manager, within which multiple calls to `contract` will share intermediate computations. This is very useful e.g. in message passing algorithms. The first draft of this PR is copied directly [from Pyro](https://github.com/uber/pyro/blob/1b1910eadff80178197879bf847beb4443fd9014/pyro/ops/einsum/shared.py), where we are using sharing for probabilistic inference in graphical models.

The implementation uses a special internal backend `opt_einsum.backends.shared`. When a sharing context is opened
```py
with shared_intermediates():
    contract(..., backend='foo')
    contract(..., backend='foo')
```
a special `handle_sharing()` context is activated inside the `contract` function, temporarily switching to the `shared` backend, and setting the original backend `foo` in a global `shared._CURRENT_BACKEND` variable. The sharing backend then performs all the original `foo` operations and also memoizes them.

One design choice is to memoize by `id()` rather than by value. This makes sense from a computational perspective (equality comparison is expensive), but requires a bit more care by users.

A second design choice is to expose the memoization `cache` to users. This makes it easy to share intermediates by passing the cache around, as is necessary when it is difficult to perform all computation in a context:
```py
with shared_intermediates() as cache:
    contract(...)
...pass control elsewhere...
with shared_intermediates(cache):  # <-- reuse previous cache
    contract(...)
del cache  # <-- intermediates will be garbage collected
```

## Todos
- [x] Cache type conversions e.g. `numpy`->`torch`->`numpy` so args can be compared by id.
- [x] fix failing tests
- [x] add tests that pass around the cache
- [x] add tests where we tests the normal ordering:
    ```py
    with shared_intermeidates() as cache:
        a = np.random.rand(4, 4)
        b = np.random.rand(4, 4)
        contract("ab,bc->ac", a, b)

    assert get_cache_key("zc,az->ac", b, a) in cache
    ...
    ```
- [x] test against the Pyro [rm-einsum-shared](https://github.com/uber/pyro/pull/1327) branch
- [x] test nested sharing
- [x] write docs
- [x] move `_alpha_canonicalize()` to `parser.py`
- [x] test caching with constant expressions

## Questions
- [x] Could a maintainer suggest where the `handle_sharing()` context should be moved? I've attempted to insert it into `ContractExpression.__call__()`, but I'm a bit lost.

## Status
- [x] Ready to go


opt_einsum/contract.py
```diff
@@ -626,28 +627,29 @@ def __call__(self, *arrays, **kwargs):
...
-            return self._contract(ops, out, backend, evaluate_constants=evaluate_constants)
-
-        except ValueError as err:
-            original_msg = str(err.args) if err.args else ""
-            msg = ("Internal error while evaluating `ContractExpression`. Note that few checks are performed"
-                   " - the number and rank of the array arguments must match the original expression. "
-                   "The internal error was: '%s'" % original_msg, )
-            err.args = msg
-            raise
+        with handle_sharing(backend) as backend:
```
fritzo: This is really the only line changed; I've simply moved the following lines inside this context.

fritzo:
cc @eb8680

jcmgray:
Nice, looking forward to testing this and putting it to use in the wild! I do have a suggestion to simplify this a bit further (sorry for suggesting all these refactors!).

Basically, instead of making it a backend, and then storing the real desired backend in a temporary and using the ``handle_sharing``, the cached versions of these could could just be turned on directly in the core ``_tensordot``, ``_einsum`` etc. Maybe as decorators defined (along with ``shared_intermediaries``) in a ``shared.py`` file.

e.g.

```python 
# opt_einsum/shared.py

def shared_intermediaries(...):
    # and other stuff

def tensordot_cache_wrap(tensordot_fn):

    def tensordot_cached(x, y, axes, backend):
        if not _SHARING_STACK:
            return tensordot_fn(x, y, axes, backend)

        ...  # parse and get global cache etc
        key = (id(x), id(y), axes, backend)
        if key not in cache:
            cache[key] = tensordot_fn(x, y, axes, backend)

        return cache[key]

    return tensordot_cached
```

The advantage is that in ``contract.py`` we then just need
```python
@tensordot_cache_wrap
def _tensordot(x, y, axes, backend)
    ....
```
I.e. the caching logic is really factored out and ``handle_shared`` and ``_CURRENT_BACKEND`` are not needed. Additionally, the calls would be explicitly routed through ``_einsum``, ``_tensordot`` & ``_transpose`` (in future they might be modified). The decorator approach is obviously optional - just a nice away of keeping the core ``contract`` logic clean -- but the key thing is that ``shared_intermediaries`` might not need to use the ``backend`` api itself.

Also the tests seem to be failing on travis currently? but otherwise looking good!

opt_einsum/backends/shared.py
```diff
@@ -0,0 +1,116 @@
+import contextlib
+import numbers
+from collections import OrderedDict
+
+from ..parser import get_symbol
+from .dispatch import get_func
+
+_SHARING_STACK = []
+_CURRENT_BACKEND = []
+
```
jcmgray: Could these just be explicit ``global`` variables? Or does the list functionality enable nested ``sharing`` with different caches or something?

opt_einsum/backends/shared.py
```diff
@@ -0,0 +1,116 @@
...
+    cache[key] = result
+    return result
+
+
+def tensordot(x, y, axes=2):
+    backend = _CURRENT_BACKEND[0]
+    cache = _SHARING_STACK[-1]
+    cache['tensor', id(x)] = x
+    cache['tensor', id(y)] = y
+
```
jcmgray: Do we need to cache the tensors on their own, are they every retrieved?

opt_einsum/backends/shared.py
```diff
@@ -0,0 +1,116 @@
...
+        cache['tensor', id(d)] = d
+
+    # compute a canonical hash, modulo commutativity
+    inputs, output = equation.split('->')
+    inputs = inputs.split(',')
+    canonical = sorted(zip(inputs, operands), key=lambda x: id(x[1]))
+    canonical_inputs = ','.join(input_ for input_, _ in canonical)
+    canonical_equation = _alpha_canonicalize('{}->{}'.format(canonical_inputs, output))
+    canonical_operands = tuple(d for _, d in canonical)
+    key = 'einsum', backend, canonical_equation, tuple(map(id, canonical_operands))
```
jcmgray: Does this bit enable it so that ``einsum('ab,bc->ca', x, y)`` matches e.g. ``einsum('jk,ij->ki', y, x)``? If so, nice!

fritzo: Yes, we're accounting for a little bit of commutativity in the cache lookup. We did this in Pyro to improve our sharing.

opt_einsum/backends/shared.py
```diff
@@ -0,0 +1,116 @@
...
+        A dictionary in which sharing results are stored. If ignored,
+        sharing results will be garbage collected when this context is
+        exited. This dict can be passed to another context to resume
+        sharing.
+    """
+    if cache is None:
+        cache = {}
+    _SHARING_STACK.append(cache)
+    yield cache
+    _SHARING_STACK.pop()
```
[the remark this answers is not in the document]
fritzo: The try-finally is automatically performed by `@contextlib.contextmanager`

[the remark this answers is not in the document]
jcmgray: Are you sure about this specifically for ``contextmanager``? The [docs](https://docs.python.org/3/library/contextlib.html) don't mention it and in my tests it is needed otherwise the post yield code is never reached. Testing with these snippets:
```python
from contextlib import contextmanager

things = []

@contextmanager
def TrySomething(value):
    things.append(value)
#     try:
    yield things[-1]
#     finally:
    things.pop() 
```
then
```python
with TrySomething('hello'):
    raise ValueError
```
gives  ``things=['hello']`` without the try/finally but ``[]`` with.

opt_einsum/backends/shared.py
jcmgray: Ah ok! I hadn't realised that about ``id``.

opt_einsum/backends/shared.py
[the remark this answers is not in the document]
fritzo: Gosh, it looks like I'm mistaken. I'll fix and push. Thanks for catching this!

opt_einsum/backends/shared.py
dgasmith: +1 for normal ordering. I think this takes care of most edge cases. The only missing edge case that I can think of is if a user does take identical views in different contexts which would require checking of the `__array_interface__` syntax. Seems like a stretch to take care of in the first pass.

opt_einsum/backends/shared.py
```diff
@@ -0,0 +1,116 @@
...
+
+
+def einsum(equation, *operands):
+    backend = _CURRENT_BACKEND[0]
+    cache = _SHARING_STACK[-1]
+    for d in operands:
+        cache['tensor', id(d)] = d
+
+    # compute a canonical hash, modulo commutativity
+    inputs, output = equation.split('->')
```
dgasmith: Are we guaranteed to have a "->" at this stage? I think so, but worth double checking.

opt_einsum/backends/shared.py
```diff
@@ -0,0 +1,116 @@
...
+        return cache[key]
+
+    result = get_func('transpose', backend)(a, axes)
+    cache[key] = result
+    return result
+
+
+def tensordot(x, y, axes=2):
+    backend = _CURRENT_BACKEND[0]
+    cache = _SHARING_STACK[-1]
```
dgasmith: Can we normal order `x, y` here as well using `id` as below? With indices explicitly labeled I think this should be ok.

opt_einsum/backends/shared.py
dgasmith: Fun example:
```
>>> a = np.random.rand(4, 4)
>>> id(a[0])
4560101008
>>> id(a[0])
4585119616
>>> id(a[0])
4585119616
>>> id(a[0])
4585119616
>>> id(a[1])
4585119616
>>> id(a[2])
4585119616
```

opt_einsum/tests/test_sharing.py
```diff
@@ -0,0 +1,166 @@
...
+            equation = '{}->{}'.format(inputs, target)
+            path_info = contract_path(equation, *xs)
+            print(path_info[1])
+            contract(equation, *xs, backend='torch')
+        num_exprs_sharing = compute_cost(cache)
+
+    print('-' * 40)
+    print('Without sharing: {} expressions'.format(num_exprs_nosharing))
+    print('With sharing: {} expressions'.format(num_exprs_sharing))
+    assert num_exprs_nosharing > num_exprs_sharing
```
dgasmith: We should consider adding additional tests where we tests the normal ordering:
```
with shared_intermeidates() as cache:
    a = np.random.rand(4, 4)
    b = np.random.rand(4, 4)
    contract("ab,bc->ac", a, b)

   assert get_cache_key("zc,az->ac", b, a) in cache
   ...
```

opt_einsum/backends/shared.py
```diff
@@ -0,0 +1,116 @@
...
+def einsum(equation, *operands):
+    backend = _CURRENT_BACKEND[0]
+    cache = _SHARING_STACK[-1]
+    for d in operands:
+        cache['tensor', id(d)] = d
+
+    # compute a canonical hash, modulo commutativity
+    inputs, output = equation.split('->')
+    inputs = inputs.split(',')
+    canonical = sorted(zip(inputs, operands), key=lambda x: id(x[1]))
```
dgasmith: It might be good to break this code out into another function. I have a feeling that tokenizing contractions will be very useful in the future.

opt_einsum/backends/shared.py
fritzo: I believe we can't normal-order `x,y` because their non-contracted dimensions are treated differently: `x`'s are on the left and `y`'s are on the right.

opt_einsum/backends/shared.py
fritzo: I believe so, but I'm not super familiar with the entire `einsum` syntax. At this stage we are guaranteed to have the same `equation` that was passed to `contract(equation, ...)`.

opt_einsum/backends/shared.py
jcmgray: One possibility would be to check for (x, y, axes) then (y, x, axes[::-1]) and if its the second just perform the transposition (which will be something like ``transpose(x, [2, 3, 4, 0, 1])``.

fritzo: Yeah, in this first PR I've implemented lookup modulo-commutativity but not modulo-transpose. I agree that we could do this modulo-transpose lookup here and also in the `einsum` part. But I'm still struggling to get tests to pass even in the current PR, so I'd be happy to limit this first PR to commutativity.

dgasmith: +1 to getting something in, can we flag this and make an issue so that we do not forget about this point?

fritzo:
> @jcmgray Maybe as decorators ...

I like this idea! I'll try refactoring and see if I have better luck getting tests to pass with the decorator version.

opt_einsum/backends/shared.py
jcmgray: Yeah absolutely - it's certainly not necessary for this PR. Good to keep track of these things so they can maybe be added them later however!

fritzo:
@jcmgray I've refactored to use decorators as you suggested, and indeed the changes are now minimally intrusive.

One issue is that I've also had to decorate `to_torch` and `to_cupy`. I don't know how to wrap the other automatically-converted backends, but all backends will work with sharing if the user does manual conversion.

jcmgray:
Yes thanks very much for that change, it's looking great. My notes at this point are completely optional extensions:

* With regards to the conversion functions, I can't think of any reason that every ``to_{backend}`` function could't be memoized. At the point it would be called ``tensorflow`` (in non-eager mode) and ``theano`` are just building the expression, so there's no direct speed advantage, but I can imagine that it only helps the compilers if it can see that some tensors are the same object. On the other hand, there is no single function ``to_backend(x, backend=...)`` at the moment so it gets a bit messy -- could well be left for the moment.

* Do we need to think about multi-threading scenarios? With one thread popping the cache early?

* On a similar note, it might be nice to test the nested sharing if it's an intentional feature.

* As I think @dgasmith pointed out, it also might be nice to just move the einsum pre-cache canonicalization into its own function in ``parser`` since it might be useful elsewhere.


But like I say, already looks good to go from my perspective without these niche things! Docs-wise, maybe added to the ``readme.md`` bullet point list and it's own page in the main docs, @dgasmith ?

fritzo:
> @jcmgray Do we need to think about multi-threading scenarios?

I really don't know, as I've never used threads in Python. What do you think?

fritzo:
Ok, I think this should be ready to go when tests pass. I can open issues for remaining improvements.

LICENSE
```diff
@@ -1,6 +1,7 @@
 The MIT License (MIT)
 
 Copyright (c) 2014 Daniel Smith
+Copyright (c) 2018 Uber Technologies
```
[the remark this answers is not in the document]
dgasmith: Yea, could we move this to `sharing.py`? We should probably look at changing the copyright to the "opt_einsum developers" in the future. I need to look into this angle of things a bit more.

docs/source/sharing_intermediates.rst
```diff
@@ -0,0 +1,36 @@
...
+    >>> marginals = {output: contract('{}->{}'.format(inputs, output), *factors)
+    >>>              for output in 'abcdef'}
+
+To share this computation, we can perform all contractions in a shared context
+
+.. code:: python
+
+    >>> with shared_intermediates():
+    >>>     marginals = {output: contract('{}->{}'.format(inputs, output), *factors)
+    >>>                  for output in 'abcdef'}
```
dgasmith: Might be worth showing a quick timing comparison on bigger tensors with an explicit demonstration of the contractions done.

opt_einsum/tests/test_sharing.py
```diff
@@ -0,0 +1,363 @@
...
+        expr(x, y, z2, backend=backend)
+        num_exprs_sharing = count_cached_ops(cache)
+
+    print('-' * 40)
+    print('Without sharing: {} expressions'.format(num_exprs_nosharing))
+    print('With sharing: {} expressions'.format(num_exprs_sharing))
+    assert num_exprs_nosharing['einsum'] > num_exprs_sharing['einsum']
+
+
+def compute_cost(cache):
```
dgasmith: Can this be `_compute_cost`?

opt_einsum/sharing.py
```diff
@@ -0,0 +1,135 @@
...
+def _save_tensors(*tensors):
+    """Save tensors in the cache to prevent their ids from being recycled.
+    This is needed to prevent false cache lookups.
+    """
+    cache = _SHARING_STACK[-1]
+    for tensor in tensors:
+        cache['tensor', id(tensor)] = tensor
+
+
+def _memoize(key, fn, *args, **kwargs):
```
dgasmith: Can we add docstrings on the next 4 functions? Not a lot, but just something to indicate their use.

dgasmith:
@jcmgray Can you give this a final review as well?

jcmgray:
> @fritzo I really don't know, as I've never used threads in Python. What do you think?

I thought about this a bit, since it's not completely unlikely that someone at some point might parallelize some numeric code which uses ``opt_einsum`` sharing in a thread pool or something. It might be inefficient but nothing catastrophic I think -> each thread will add its own cache to the list, but only use whichever is last (so a bit randomly jumbled). Additionally, while *any* one thread is in a shared context, ``opt_einsum`` in *all* threads will cache intermediaries.

Anyway I can't think of any remedy so really just noting it! I guess it's just a minor cost of the major convenience here of using module global state.

NB. if you supply the same cache to all threads, they will all add it, but will all use it, so that sidesteps the inefficiency to an extent.

opt_einsum/sharing.py
fritzo: done.

fritzo:
Ok, I've addressed all review comments.

@dgasmith and @jcmgray thank you for your detailed review!

dgasmith:
Everything looks good to me. Thanks for the great PR! We will look at releasing a 2.2 soon to get these changes into production.

jcmgray:
This is a really cool addition, thanks @fritzo!

dgasmith:
I usually do for larger projects, but `opt_einsum` is small enough that there are not a ton of commits coming in and we haven't declared a git model that we use.

PR #48, opened 2018-08-23
Make get_symbol(large_number) work in Python2

### LINKED ISSUE #42: Support for get_symbol(116) in Python 2
`get_symbol(x)` currently fails in Python 2 for any `x >= 116`:
```
i = 116

    def get_symbol(i):
        """Get the symbol corresponding to int ``i`` - runs through the usual 52
        letters before resorting to unicode characters, starting at ``chr(192)``.

        Examples
        --------
        >>> get_symbol(2)
        'c'

        >>> oe.get_symbol(200)
        'Ŕ'

        >>> oe.get_symbol(20000)
        '京'
        """
        if i < 52:
            return einsum_symbols_base[i]
>       return chr(i + 140)
E       ValueError: chr() arg not in range(256)
```
Can we fix this by use `unichr()` and unicode strings in Python 2?

fritzo:
Fixes #42

## Description
This provides a compatibility shim for Python 2 so that `get_symbol(n)` can work for larger n.

Note that Python 2's `numpy.einsum` still does not accept unicode. This means that, while `opt_einsum` can now work with arbitrarily large contractions in Python 2, it only works with contractions whose individual calls to Numpy involve fewer than 52 (I believe) characters.

## Tested

- [x] added a regression test
- [x] enabled previously skipped tests
- [x] tested in a Pyro application

## Status
- [x] Ready to go

### DIFF
--- a/opt_einsum/sharing.py
+++ b/opt_einsum/sharing.py
@@ -129,7 +129,7 @@ def einsum_cache_wrap(einsum):
         canonical = sorted(zip(inputs, map(id, operands)), key=lambda x: x[1])
         canonical_ids = tuple(id_ for _, id_ in canonical)
         canonical_inputs = ','.join(input_ for input_, _ in canonical)
-        canonical_equation = alpha_canonicalize('{}->{}'.format(canonical_inputs, output))
+        canonical_equation = alpha_canonicalize(canonical_inputs + "->" + output)
         key = 'einsum', backend, canonical_equation, canonical_ids
         return _memoize(key, einsum, equation, *operands, backend=backend)
 
--- a/opt_einsum/tests/test_contract.py
+++ b/opt_einsum/tests/test_contract.py
@@ -145,7 +145,6 @@ def test_compare_blas(string):
     assert np.allclose(ein, opt)
 
 
-@pytest.mark.skipif(sys.version_info[0] < 3, reason='requires python3')
 @pytest.mark.parametrize("string", tests)
 def test_compare_blas_greek(string):
     views = helpers.build_views(string)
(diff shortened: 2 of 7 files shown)


dgasmith:
There are also tests [here](https://github.com/dgasmith/opt_einsum/blob/master/opt_einsum/tests/test_input.py#L233) and [here](https://github.com/dgasmith/opt_einsum/blob/master/opt_einsum/tests/test_contract.py#L120) which can likely be fixed by these changes.

opt_einsum/tests/test_contract.py
```diff
@@ -117,15 +117,14 @@ def test_drop_in_replacement(string):
     assert np.allclose(opt, np.einsum(string, *views))
 
 
-@pytest.mark.skipif(sys.version_info[0] < 3, reason='requires python3')
 @pytest.mark.parametrize("string", tests)
```
dgasmith: Thanks, I think there are two more in this file. I think this is good to go after those are patched?

fritzo: Great, I believe this is good to go now. Looking forward to using this in Pyro.

opt_einsum/sharing.py
```diff
@@ -129,7 +129,7 @@ def cached_einsum(*args, **kwargs):
         canonical = sorted(zip(inputs, map(id, operands)), key=lambda x: x[1])
         canonical_ids = tuple(id_ for _, id_ in canonical)
         canonical_inputs = ','.join(input_ for input_, _ in canonical)
-        canonical_equation = alpha_canonicalize('{}->{}'.format(canonical_inputs, output))
```
dgasmith: Any reason to avoid `format` here?

fritzo:
@dgasmith Thanks for the speedy response!

PR #49, opened 2018-08-23
tiny readme fix

jcmgray:
Just a spelling fix for the readme. Also it irked me that a single bullet point spilled onto a second line by a single word so I've 'fixed' that as well.


dgasmith:
Ah, good catch-- thanks!

PR #53, opened 2018-08-24
make shared_intermediates thread-safe

### LINKED ISSUE #51: Support multi-threading in shared_intermediates
Following up on https://github.com/dgasmith/opt_einsum/pull/43#issuecomment-414661988, the `shared_intermediates` cache is currently not locked.

One option would be to assign the thread id of a unique owner of the cache, such that at least thread conflicts could be detected and reported, rather than mere silent performance degradation.

jcmgray:
## Description
Makes ``shared_intermediates`` thread safe, by storing the caches in a dict referenced by thread id. All threads can (still) share the same cache among themselves, if it is supplied to them all. Resolves #51. cc @fritzo.

## Todos
  - [x] Implement a python2/3 compatible test

## Status
- [x] Ready to go

### DIFF
--- a/opt_einsum/sharing.py
+++ b/opt_einsum/sharing.py
@@ -3,36 +3,41 @@
 import contextlib
 import functools
 import numbers
-from collections import Counter
+from collections import Counter, defaultdict
 import threading
 
 from .parser import alpha_canonicalize, parse_einsum_input
 
 
-_SHARING_STACK = {}
+_SHARING_STACK = defaultdict(list)
+
+
+try:
+    get_thread_id = threading.get_ident
+except AttributeError:
+    def get_thread_id():
+        return threading.current_thread().ident
 
 
 def currently_sharing():
     """Check if we are currently sharing a cache -- thread specific.
     """
-    return threading.get_ident() in _SHARING_STACK
+    return get_thread_id() in _SHARING_STACK
 
 
 def get_sharing_cache():
     """Return the most recent sharing cache -- thread specific.
     """
-    return _SHARING_STACK[threading.get_ident()][-1]
+    return _SHARING_STACK[get_thread_id()][-1]
 
 
 def _add_sharing_cache(cache):
-    tid = threading.get_ident()
-    _SHARING_STACK.setdefault(tid, [])
-    _SHARING_STACK[tid].append(cache)
+    _SHARING_STACK[get_thread_id()].append(cache)
 
 
-def _remove_sharing_cache(cache):
-    tid = threading.get_ident()
-    _SHARING_STACK[tid].remove(cache)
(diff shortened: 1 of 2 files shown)


opt_einsum/sharing.py
```diff
@@ -4,10 +4,37 @@
...
 
-_SHARING_STACK = []
+
+_SHARING_STACK = {}
+
+
+def currently_sharing():
+    """Check if we are currently sharing a cache -- thread specific.
+    """
+    return threading.get_ident() in _SHARING_STACK
```
fritzo: Hmm in Python 2, `threading.get_ident` does not exist

jcmgray: Yes, just realized that! I'll have a quick look for an alternative.

opt_einsum/sharing.py
```diff
@@ -4,10 +4,37 @@
 import functools
 import numbers
 from collections import Counter
+import threading
 
 from .parser import alpha_canonicalize, parse_einsum_input
 
-_SHARING_STACK = []
+
+_SHARING_STACK = {}
```
fritzo: Consider making this a `defaultdict(list)` to avoid the `setdefault` complexity below?

opt_einsum/sharing.py
```diff
@@ -4,10 +4,37 @@
...
+
+def _add_sharing_cache(cache):
+    tid = threading.get_ident()
+    _SHARING_STACK.setdefault(tid, [])
+    _SHARING_STACK[tid].append(cache)
+
+
+def _remove_sharing_cache(cache):
+    tid = threading.get_ident()
+    _SHARING_STACK[tid].remove(cache)
```
fritzo: What is the advantage of `.remove()` over `.pop()`? If these are really stacks, then we should use `.pop()`. If these are not really stacks, then we should use sets or something.

opt_einsum/sharing.py
jcmgray: Yes will do.

opt_einsum/sharing.py
jcmgray: Yes should be exactly the same for the current behaviour, ``remove(cache)`` felt more explicit but is probably unnecessary, I'll update it.

jcmgray:
Thanks for those comments @fritzo! I've made those changes and added a test compatible with both py2/py3. Ready to go from my end.

jcmgray:
It doesn't happen automatically, but threads sharing a cache is already possible if you explicitly supply the cache to use (they'll each add it their own 'stack', but will all still be referencing the same object).

dgasmith:
Good point, currently doing a docs pass and I will add that.

PR #58, opened 2018-08-28
allow sharing for tensorflow and theano as numpy backends

### LINKED ISSUE #50: Support more backends in shared_intermediates
Following up on https://github.com/dgasmith/opt_einsum/pull/43#issuecomment-414661988, it would be nice if `with shared_intermediates():` supported more backends. This could start by refactoring towards a common `to_backend` function that is memoized.

## Status

- [x] numpy
- [x] torch
- [x] cupy
- [ ] tensorflow
- [ ] theano
- [ ] sparse

jcmgray:
## Description
Allows sharing to be used with tensorflow and theano as numpy backends. Closes #50. The ``to_{backend}`` functions are all called internally within their respective ``{backend.py}`` submodules so I didn't try and factorize out a new ``to_backend`` function. I think it will be pretty easy (and rare) to add new conversion backends anyway. Also the tests are in ``test_backends.py`` rather than ``test_sharing.py`` just because the tensorflow tests require a ``session``, and the theano tests can't use ``einsum`` itself. 

## Status
- [x] Ready to go

### DIFF
--- a/docs/source/index.rst
+++ b/docs/source/index.rst
@@ -7,7 +7,10 @@ dimension and index. However, it is only optimized to contract two terms
 at a time resulting in non-optimal scaling.
 
 For example, consider the following index transformation:
-``M_{pqrs} = C_{pi} C_{qj} I_{ijkl} C_{rk} C_{sl}``
+
+.. math::
+
+    M_{pqrs} = C_{pi} C_{qj} I_{ijkl} C_{rk} C_{sl}
 
 Consider two different algorithms:
 
--- a/opt_einsum/backends/tensorflow.py
+++ b/opt_einsum/backends/tensorflow.py
@@ -6,7 +6,10 @@ from __future__ import absolute_import
 
 import numpy as np
 
+from ..sharing import to_backend_cache_wrap
+
 __all__ = ["to_tensorflow", "build_expression", "evaluate_constants"]
+
 
 _CACHED_TF_DEVICE = None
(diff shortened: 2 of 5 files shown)


docs/source/index.rst
```diff
@@ -7,7 +7,10 @@ dimension and index. However, it is only optimized to contract two terms
 at a time resulting in non-optimal scaling.
 
 For example, consider the following index transformation:
-``M_{pqrs} = C_{pi} C_{qj} I_{ijkl} C_{rk} C_{sl}``
+
+.. math::
+
+    M_{pqrs} = C_{pi} C_{qj} I_{ijkl} C_{rk} C_{sl}
```
jcmgray: Ah this has crept in as a small docs change. Can remove if necessary @dgasmith.

jcmgray:
Yes that would have been nice but I've ended up adding them to [``test_backends.py``](https://github.com/dgasmith/opt_einsum/blob/3d72bba4087a56411b19ce30c1a68a70773113dd/opt_einsum/tests/test_backends.py#L7) tests for two reasons: 

1. tensorflow requires a ``Session`` object to be setup and torn down between runs (which is defined with a special config in ``test_backends.py``.
2. theano doesn't provide ``einsum`` and so can't actually perform many of the test contractions in the ``test_sharing.py``

I didn't think too hard about unifying the tests but it is probably possible.

Issue #56, opened 2018-08-27
Reduce greedy complexity to sub-quadratic?


fritzo:
The greedy strategy can be quite slow when given hundreds of tensors. Is it possible to reduce its complexity to sub-quadratic, possibly by:
- using incremental computations when possible
- using memoization when possible
- maintaining a priority queue rather than recomputing min-cost contraction at each step

jcmgray:
It would certainly nice to speed up greedy further. Just a overview of the current algorithm:

1. Initially assess all inner contractions (i.e. those with shared indices) of ``n`` tensors. If the contraction graph has regularity ``k`` (~ average number of indices per tensor) this has cost propto ``nk`` (ignoring the cost of filtering for inner contractions which is pretty negligible). 
2. Find the minimum cost contraction (time spent doing ``min`` is currently relatively negligible also) and add that contraction to the contraction list.
3. Modify the other possible contractions that are still valid with the new result, small  cost but propto ``n`` and have to this for each of ``n`` steps.
4. Compute the approximately ``k`` new possible contractions using the new result - also for each step. With this method the same contraction should never be recomputed.

So overall the main steps of the algorithm take ``nk`` time (don't think we could do better than this), but the steps involving modifying the possible contractions take more like ``n^2`` time. Here's some profiling of a very large contraction ([generated using this gist](https://gist.github.com/jcmgray/2fb88f1b8473eef5070316f02e26d66e)):

![screenshot from 2018-08-27 23-50-44](https://user-images.githubusercontent.com/8982598/44690811-1c165300-aa54-11e8-801e-73bebb528370.png)

From that I think the biggest savings will had updating the internal format used to represent contractions. Including:

* storing the tensors in a dict (keyed by indices) rather than a list so that we don't have to manipulate index positions.
* not storing the full output set with every potential contraction (which then have to be modified). This is quite natural for small contractions, but when each of 1000 contractions is storing its 999 output index sets gets a bit inefficient.

 There might be other benefits to these kind of changes (like easier manipulation of the path once found, e.g. to ensure all constant operations are at the beginning) but it might also be quite an involved change.

fritzo:
I've prototyped a cheaper path optimizer in https://github.com/uber/pyro/pull/1424. It is still quadratic, but is more than 10x faster on large paths. I'd like to try to further reduce its complexity, then move some of the patterns upstream.

fritzo:
[...]

PR #60, opened 2018-09-30
Add prototype cheap optimizer

fritzo:
Addresses #56 
Mirrors ~~https://github.com/uber/pyro/pull/1424~~ https://github.com/uber/pyro/pull/1430

## Description

Implements a cheap approximate optimizer for very large paths. This is currently over 100x faster on some large contractions in Pyro, but produces slightly suboptimal paths.

## Todos
- [x] Fix issues with determinism, required by `shared_intermediates`
- [x] Ensure algorithm is not much worse than `greedy`
- ~~Add quality tests (will be easier after #61)~~

## Questions
- [x] How should we be computing the final outer products? Can we do this all in one go?
- [x] What tests should I add?

## Status
- [x] Ready to go

### DIFF
--- a/opt_einsum/contract.py
+++ b/opt_einsum/contract.py
@@ -99,7 +99,7 @@ def contract_path(*operands, **kwargs):
         - 'optimal' An algorithm that tries all possible ways of
           contracting the listed tensors. Scales exponentially with
           the number of terms in the contraction.
-        - 'cheap' An approximate algorithm that is cheaper than 'greedy' for
+        - 'eager' An approximate algorithm that is cheaper than 'greedy' for
           very large contractions.
         - 'auto' Use the optimal approach for a small number of terms (currently
           4 or less), else use the greedy approach.
@@ -255,8 +255,8 @@ def contract_path(*operands, **kwargs):
         path = paths.optimal(input_sets, output_set, dimension_dict, memory_arg)
     elif path_type in ("greedy", "opportunistic", "auto"):
         path = paths.greedy(input_sets, output_set, dimension_dict, memory_arg)
-    elif path_type == 'cheap':
-        path = paths.cheap(input_sets, output_set, dimension_dict)
+    elif path_type == 'eager':
+        path = paths.eager(input_sets, output_set, dimension_dict)
     else:
         raise KeyError("Path name '{}' not found".format(path_type))
 
@@ -394,7 +394,7 @@ def contract(*operands, **kwargs):
         - 'optimal' An algorithm that tries all possible ways of
           contracting the listed tensors. Scales exponentially with
           the number of terms in the contraction.
-        - 'cheap' An approximate algorithm that is cheaper than 'greedy'
+        - 'eager' An approximate algorithm that is cheaper than 'greedy'
           for very large contractions.
 
     memory_limit : {None, int, 'max_input'} (default: None)
--- a/opt_einsum/contract.py
+++ b/opt_einsum/contract.py
@@ -42,23 +42,23 @@ class PathInfo(object):
         header = ("scaling", "BLAS", "current", "remaining")
 
         path_print = [
-            "  Complete contraction:  {}\n".format(overall_contraction),
-            "         Naive scaling:  {}\n".format(len(self.indices)),
-            "     Optimized scaling:  {}\n".format(max(self.scale_list)),
-            "      Naive FLOP count:  {:.3e}\n".format(naive_cost),
-            "  Optimized FLOP count:  {:.3e}\n".format(opt_cost),
-            "   Theoretical speedup:  {:3.3f}\n".format(speedup),
-            "  Largest intermediate:  {:.3e} elements\n".format(largest_intermediate),
-            "-" * 80 + "\n",
-            "{:>6} {:>11} {:>22} {:>37}\n".format(*header),
-            "-" * 80
+            u"  Complete contraction:  {}\n".format(overall_contraction),
+            u"         Naive scaling:  {}\n".format(len(self.indices)),
+            u"     Optimized scaling:  {}\n".format(max(self.scale_list)),
+            u"      Naive FLOP count:  {:.3e}\n".format(naive_cost),
+            u"  Optimized FLOP count:  {:.3e}\n".format(opt_cost),
+            u"   Theoretical speedup:  {:3.3f}\n".format(speedup),
+            u"  Largest intermediate:  {:.3e} elements\n".format(largest_intermediate),
+            u"-" * 80 + "\n",
+            u"{:>6} {:>11} {:>22} {:>37}\n".format(*header),
+            u"-" * 80
         ]
 
         for n, contraction in enumerate(self.contraction_list):
             inds, idx_rm, einsum_str, remaining, do_blas = contraction
             remaining_str = ",".join(remaining) + "->" + self.output_subscript
             path_run = (self.scale_list[n], do_blas, einsum_str, remaining_str)
-            path_print.append("\n{:>4} {:>14} {:>22} {:>37}".format(*path_run))
+            path_print.append(u"\n{:>4} {:>14} {:>22} {:>37}".format(*path_run))
 
         return "".join(path_print)
 
--- a/opt_einsum/contract.py
+++ b/opt_einsum/contract.py
@@ -35,6 +35,8 @@ def contract_path(*operands, **kwargs):
         - 'optimal' An algorithm that tries all possible ways of
           contracting the listed tensors. Scales exponentially with
           the number of terms in the contraction.
+        - 'cheap' An approximate algorithm that is cheaper than 'greedy' for
+          very large contractions.
         - 'auto' Use the optimal approach for a small number of terms (currently
           4 or less), else use the greedy approach.
 
@@ -194,6 +196,8 @@ def contract_path(*operands, **kwargs):
         path = paths.optimal(input_sets, output_set, dimension_dict, memory_arg)
     elif path_type in ("greedy", "opportunistic", "auto"):
         path = paths.greedy(input_sets, output_set, dimension_dict, memory_arg)
+    elif path_type == 'cheap':
+        path = paths.cheap(input_sets, output_set, dimension_dict)
     else:
         raise KeyError("Path name %s not found" % path_type)
 
@@ -349,6 +353,8 @@ def contract(*operands, **kwargs):
         - 'optimal' An algorithm that tries all possible ways of
           contracting the listed tensors. Scales exponentially with
           the number of terms in the contraction.
+        - 'cheap' An approximate algorithm that is cheaper than 'greedy'
+          for very large contractions.
 
     memory_limit : int or None (default : None)
         The upper limit of the size of tensor created, by default, this will be
--- a/opt_einsum/paths.py
+++ b/opt_einsum/paths.py
@@ -289,10 +289,10 @@ def ssa_to_linear(ssa_path):
         >>> ssa_to_linear([(0, 3), (1, 4), (2, 5)])
         [(0, 3), (1, 2), (0, 1)]
     """
-    ids = np.arange(sum(map(len, ssa_path)), dtype=np.int64)
+    ids = np.arange(1 + max(map(max, ssa_path)), dtype=np.int32)
     path = []
-    for i, ssa_ids in enumerate(ssa_path):
-        path.append(tuple(ids[ssa_id] for ssa_id in ssa_ids))
+    for ssa_ids in ssa_path:
+        path.append(tuple(int(ids[ssa_id]) for ssa_id in ssa_ids))
         for ssa_id in ssa_ids:
             ids[ssa_id:] -= 1
     return path
@@ -451,12 +451,11 @@ def _ssa_optimize(inputs, output, sizes):
 
 def eager(inputs, output, idx_dict):
     """
-    Finds the path by a quick-and-dirty method.
+    Finds the path by a three stage algorithm:
 
     1. Eagerly compute Hadamard products.
     2. Greedily compute contractions to maximize ``removed_size``
     3. Greedily compute outer products.
-    4. Finally call einsum with a single arg to fix output shape.
 
     This algorithm scales quadratically with respect to the
     maximum number of elements sharing a common dim.
--- a/opt_einsum/paths.py
+++ b/opt_einsum/paths.py
@@ -318,12 +318,12 @@ def linear_to_ssa(path):
     return ssa_path
 
 
-def _get_candidate(output, sizes, remaining, costs, dim_ref_counts, k1, k2):
+def _get_candidate(output, sizes, remaining, footprints, dim_ref_counts, k1, k2):
     either = k1 | k2
     two = k1 & k2
     one = either - two
     k12 = (either & output) | (two & dim_ref_counts[3]) | (one & dim_ref_counts[2])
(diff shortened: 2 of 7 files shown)


opt_einsum/paths.py
```diff
@@ -272,3 +279,199 @@ def greedy(input_sets, output_set, idx_dict, memory_limit):
...
+                dim_to_keys[dim].add(k12)
+        remaining[k12] = next(ssa_ids)
+        _update_ref_counts(dim_to_keys, dim_ref_counts, k1 | k2 - output)
+
+        # Find new candidate contractions.
+        k1 = k12
+        k2s = set(k2 for dim in k1 for k2 in dim_to_keys[dim] if k2 != k1)
+        _push_candidate(output, sizes, remaining, footprints, dim_ref_counts, k1, k2s, queue)
+
+    # Greedily compute pairwise outer products.
```
fritzo: @jcmgray Am I doing this right? I don't really understand the outer product phase, and I'm not sure what the difference is between this and the simpler
```py
# Perform a final outer-product contraction.
ssa_path.append(tuple(remaining.values()))
return ssa_path
```

jcmgray: The difference is just efficiency, although I agree it's quite a niche scenario. Doing pairwise outer products is generally faster than performing the entire outer product simultaneously with einsum (and whats more permits non-einsum backends). There's a test for this: ``test_can_optimize_outer_products``. 

I think you should be able to use the exact same costing/logic as the inner products, the only difference being that we only invoke checking all the outer products if there >1 remaining tensors but no remaining inner products (this is an approximate heuristic - sometimes the optimal path involves performing outer products first).

opt_einsum/paths.py
```diff
@@ -272,3 +279,199 @@ def greedy(input_sets, output_set, idx_dict, memory_limit):
...
+        count: set(dim for dim, keys in dim_to_keys.items() if len(keys) >= count) - output
+        for count in [2, 3]}
+
+    # Find initial candidate contractions.
+    queue = []
+    for dim, keys in dim_to_keys.items():
+        keys = list(keys)
+        for i, k1 in enumerate(keys):
+            k2s = keys[:i]
+            _push_candidate(output, sizes, remaining, footprints, dim_ref_counts, k1, k2s, queue)
```
jcmgray: Still wrapping my head around this part a bit. Could you not just iterate over each ``dim`` that appears more than twice (``dim_ref_counts[2]``)?

I think I see that that you only ever need to add the minimum cost from ``k1`` between ``k2s``, is this the reason for not just pushing all the pairwise candidates?

opt_einsum/paths.py
```diff
@@ -272,3 +279,199 @@ def greedy(input_sets, output_set, idx_dict, memory_limit):
...
+            del linear_to_ssa[id_]
+        linear_to_ssa.append(next(new_ids))
+    return ssa_path
+
+
+def _get_candidate(output, sizes, remaining, footprints, dim_ref_counts, k1, k2):
+    either = k1 | k2
+    two = k1 & k2
+    one = either - two
+    k12 = (either & output) | (two & dim_ref_counts[3]) | (one & dim_ref_counts[2])
```
jcmgray: I like this way of keeping track of which indices should be maintained, just a note that it would be good to use the same logic in `optimal` eventually.

opt_einsum/paths.py
```diff
@@ -272,3 +279,199 @@ def greedy(input_sets, output_set, idx_dict, memory_limit):
...
+    while queue:
+        _, ssa_id2, k2 = heapq.heappop(queue)
+        ssa_path.append((ssa_id1, ssa_id2))
+        k12 = (k1 | k2) & output
+        cost = len(k12)
+        ssa_id12 = next(ssa_ids)
+        _, ssa_id1, k1 = heapq.heappushpop(queue, (cost, ssa_id12, k12))
+
+    # Perform one final reduction to match output shape.
+    ssa_path.append((ssa_id1,))
```
jcmgray: Yes see comment above about output indices - I think this shouldn't be necessary.

jcmgray:
I've updated a gist I have of contraction finding time [here](https://gist.github.com/jcmgray/2fb88f1b8473eef5070316f02e26d66e), it looks good for large contractions!

index.

In a 'real-world' contraction, there is a slight difference in timings, maybe due to ignoring FLOP cost? Might also just be random though.

opt_einsum/paths.py
fritzo: Originally I pushed all the candidates, but ended up with something like 30000 candidates in my test case. This ended up slowing down the O(n log(n)) heap logic. Switching to eager O(n) minimization gave me about a 10x speedup on my large test case.

opt_einsum/paths.py
fritzo: Yeah `_get_candidate()` is the hotspot in this algorithm, actually this line was the hottest spot. I got about 2x speedup by switching from for loops to `frozenset` logic here.

opt_einsum/paths.py
```diff
@@ -272,3 +279,199 @@ def greedy(input_sets, output_set, idx_dict, memory_limit):
...
+
+def _ssa_optimize(inputs, output, sizes):
+    """
+    This has an interface similar to :func:`cheap` but produces a path with
+    static single assignment ids rather than recycled linear ids.
+    SSA ids are cheaper to work with and easier to reason about.
+    """
+    # A dim common to all terms might as well be an output dim.
+    inputs = list(map(frozenset, inputs))
+    output = frozenset(output) | frozenset.intersection(*inputs)
```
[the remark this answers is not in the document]
fritzo: I'll expand on this comment. This is purely for performance. The issue is that the overall algorithm is quadratic in the number of tensors that share any dimension. Often all tensors share a batch dimension. But the path we produce should be the same whether or not the problem is batched, so I assume wlog that we're not reducing along this dimension. This led to a >10x speedup in my large test case.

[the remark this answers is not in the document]
jcmgray: Ah I think I understand now - so if you don't do this, any broadcast indices make the graph maximally connected. I don't generally have any broadcast indices in my contractions thus this is a bit of an alien situation. If you have a typical one to hand as an example that might be nice for me to reason/experiment with but no worries if not!

opt_einsum/paths.py
fritzo: Great, I've added tests cases to `test_can_optimize_outer_products`, refined the objective function, and removed the extraneous final contraction (moved to top of function as an edge case).

> use the exact same costing/logic as the inner products

Note the one difference: this phase optimizes for small inputs `(x1, x2)` independently, rather than optimizing for small outputs `x12`. By using a separable objective function, this phase has linear complexity; whereas if we used the same logic as for contractions this phase would have quadratic complexity. I'm not sure how much this matters in practice.

opt_einsum/paths.py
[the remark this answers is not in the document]
fritzo: Yes this is a very common example in Pyro. Here is a large [test case](https://github.com/uber/pyro/blob/9a10d0e/tests/ops/einsum/test_paths.py#L171). Alternatively, simply append `z` to all inputs in any of your standard test cases :smile:

opt_einsum/tests/test_paths.py
```diff
@@ -172,10 +176,12 @@ def test_greedy_edge_cases():
...
 
 
-def test_can_optimize_outer_products():
+@pytest.mark.parametrize("optimize", ['greedy', 'cheap'])
+def test_can_optimize_outer_products(optimize):
     a, b, c = [np.random.randn(10, 10) for _ in range(3)]
     d = np.random.randn(10, 2)
-    assert oe.contract_path("ab,cd,ef,fg", a, b, c, d, path='greedy')[0] == [(2, 3), (0, 2), (0, 1)]
+    actual = oe.contract_path("ab,cd,ef,fg", a, b, c, d, path=optimize)[0]
+    assert actual in [[(2, 3), (0, 2), (0, 1)], [(2, 3), (0, 2), (0, 1)]]
```
dgasmith: The righthand side elements look identical here?

opt_einsum/paths.py
```diff
@@ -272,3 +279,201 @@ def greedy(input_sets, output_set, idx_dict, memory_limit):
...
+        linear_to_ssa.append(next(new_ids))
+    return ssa_path
+
+
+def _get_candidate(output, sizes, remaining, dim_ref_counts, k1, k2):
+    either = k1 | k2
+    two = k1 & k2
+    one = either - two
+    k12 = (either & output) | (two & dim_ref_counts[3]) | (one & dim_ref_counts[2])
+    cost = helpers.compute_size_by_dict(k12, sizes)
```
jcmgray: What was the reasoning for moving away from ``size(k12) - size(k1) - size(k2)``? It produces noticeable worse contractions for me in quite a few examples.

opt_einsum/paths.py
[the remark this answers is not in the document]
jcmgray: Thanks, and yes of course - but it wouldn't be realistic :wink:.

opt_einsum/paths.py
jcmgray: Nice. Yeah if you are only left with outer products there should only be a few of them since they produce an exponentially large tensor (the most extreme case for example would be about <~30 tensors each of size 2). Your approach looks good however!

opt_einsum/paths.py
fritzo: Oh ok, I'll change back to a variant of `size12 - size1 - size2`. My motivation was that this objective led to more determinism in a large example I was looking at, and the increased sharing led to overall cheaper cost. I'll see if I can achieve the same effect with a different tie-breaker.

opt_einsum/paths.py
fritzo: So I just realized that this is an important case, e.g. if we have many pairs of contractions all lumped into a single einsum, e.g. "a,a,b,b,c,c,d,d,e,e". This results in linearly-many outer products (i.e. multiplies). So it's at least a little important to avoid quadratic growth here.

opt_einsum/paths.py
jcmgray: Yes for most 'difficult' examples I've tried (e.g. contracting quantum circuits), doing it this way produces contractions with a bit worse to a lot worse scaling so my preference would be to keep it.

On a side note, it might be nice at some point to be able to supply a custom function here (e.g. just ``size12``). I particularly like the idea of supplying several even intentionally random costs and then finding the best scaling across several attempts. Might be an issue best left for later however!

dgasmith: Echoing @jcmgray here, if this isn't too much of a hotspot `size(k12) - size(k1) - size(k2)` should prevent some obviously poor pathways.

opt_einsum/paths.py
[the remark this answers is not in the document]
dgasmith: Did you find that `frozenset` was more performant than set overall? We have a number of places where we could replace `set` with `frozenset` if true.

jcmgray:
Do we want this to replace the current ``'greedy'`` method? I think it probably could, and as long as it maintains the ``size12 - size1 - size2`` cost it should find essentially the same contraction paths (modulo a bit of variance from choosing different branches when several have the same cost).

The one remaining difference (I think?) is that the current 'greedy' breaks ties using the flop-cost, but adding this back in might make most sense using a custom cost function like proposed in #64.

We could keep the current greedy around as ``'greedy_legacy'`` or something just for reference.

opt_einsum/contract.py
```diff
@@ -349,6 +353,8 @@ def contract(*operands, **kwargs):
         - 'optimal' An algorithm that tries all possible ways of
           contracting the listed tensors. Scales exponentially with
           the number of terms in the contraction.
+        - 'cheap' An approximate algorithm that is cheaper than 'greedy'
```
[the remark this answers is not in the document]
fritzo: Ha ha no love fore "cheap" :smile: Sure, I'll change to "eager".

opt_einsum/paths.py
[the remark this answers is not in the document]
fritzo: I think `frozenset` is equally performant with `set`. What I found was that `frozenset` was more performant than manually computing intersections via a for loop. The main reason I'm using `frozenset`s here is to be able to use the sets as keys of dictionaries. But I also find immutable data structures make it easier to reason about code (tuples over lists, frozensets over sets).

fritzo:
Ok, I've addressed review comments. The current cost function is `size12 - (size1 + size2) / 2`, since this seems to work on the few sharing examples I've tried. I don't mind if you change it back to `size12 - size1 - size2`.

I think it's reasonable to merge (pending further comments) and then build some experimentation tooling to determine what objective function works will. One thing that would be handy would be a FLOP counter that worked with sharing; currently I'm running an actual computation and counting number of things in the sharing `cache`.

opt_einsum/paths.py
```diff
@@ -272,3 +279,210 @@ def greedy(input_sets, output_set, idx_dict, memory_limit):
...
+        linear_to_ssa.append(next(new_ids))
+    return ssa_path
+
+
+def _get_candidate(output, sizes, remaining, costs, dim_ref_counts, k1, k2):
+    either = k1 | k2
+    two = k1 & k2
+    one = either - two
+    k12 = (either & output) | (two & dim_ref_counts[3]) | (one & dim_ref_counts[2])
+    cost = costs[k1] + costs[k2] + helpers.compute_size_by_dict(k12, sizes)
```
jcmgray: I think I would prefer two things here:

1. ``size12 - size1 - size2`` I still think is the best default, it works best for all the tricky contractions I have (current version with ``(size1 -size2) / 2`` varies between 1-4x worse), and is quite natural since it is just proportional to the amount of memory removed.

2. Maybe this was done for efficiency, but I think its a bit confusing having the cost logic spread out over several places, i.e. ``costs[k1]`` and ``costs[k1]`` are halved and negative here, but that's not very clear. If the line was explicitly ``cost = footprint[k12] - footprint[1] - footprint[2]`` I think that would be beneficial, particularly if it comes to implementing a custom cost function.

fritzo: Done, thanks for reviewing. I believe this should now be very close to the `greedy` implementation.

opt_einsum/contract.py
```diff
@@ -42,23 +42,23 @@ def __repr__(self):
...
-            "         Naive scaling:  {}\n".format(len(self.indices)),
-            "     Optimized scaling:  {}\n".format(max(self.scale_list)),
-            "      Naive FLOP count:  {:.3e}\n".format(naive_cost),
-            "  Optimized FLOP count:  {:.3e}\n".format(opt_cost),
-            "   Theoretical speedup:  {:3.3f}\n".format(speedup),
-            "  Largest intermediate:  {:.3e} elements\n".format(largest_intermediate),
-            "-" * 80 + "\n",
-            "{:>6} {:>11} {:>22} {:>37}\n".format(*header),
-            "-" * 80
+            u"  Complete contraction:  {}\n".format(overall_contraction),
```
fritzo: This fixes a Python 2 unicode bug in `PathInfo`. I could pull this out into a separate PR if you want.

jcmgray: Fine at least for me to leave this in, thanks for fixing it.

jcmgray:
All looks great to me. I've updated this gist of [timings and other tests here](https://gist.github.com/jcmgray/2fb88f1b8473eef5070316f02e26d66e) in case its of interest (main timings similar to above still).

fritzo:
@jcmgray @dgasmith Thanks for the extensive review!

Issue #55, opened 2018-08-26
Make no memory limit the default?


jcmgray:
This is more a open question than a concrete suggestion. I'm just wondering if limiting the contraction memory to the largest input is the sensible choice for most use cases? I always turn it off, and need to to see best speedups for my use cases. In the contractions where memory *is* going to be an issue, limiting the possibly contractions results in exponentially slow progress anyway.

I'm just thinking that maybe a lot of contractions could be made faster by default for users not aware of this.

Anyway, I realise my use-cases are not everyones! And that this also might have also been influenced a bit by keeping maximum compatibility with numpy's einsum.

dgasmith:
I think that your right, it doesn't make as much sense for this repository (NumPy probably still does). One drawback is that the cost of `optimal` will rise as we use the memory limit to sieve out early paths that have a very large cost. A couple thoughts:
 - We might want to look back into the optimal algorithm to see how we can still sieve bad paths, perhaps `if current_path_cost > naive cost` as a starting point.
 - Do we want to limit maximum memory in any way? Say current unused memory on the system.

jcmgray:
* Yes, some heuristic path trimming approach might be nice. See [this paper](https://arxiv.org/abs/1304.6112) where it seems an optimal, exhaustive approach is possible for up to 25 tensors at once.

* Functionally, for me, it would be nice if ``opt_einsum`` just emitted a warning when a found path used more memory than was available. But not all backends use the same memory (dask-> disk, others-> GPU) so there's no real catch-all. You could have ``memory_limit={'warn', 'system', -1, int}`` or something.

dgasmith:
Thinking back on this, it seems most of the concerns over limiting this functionality have not really come to fruition. I really like the `warn` idea overall.

jcmgray:
Shall I leave this open in case we want to think about the ``'warn'`` and ``'system'`` options?

dgasmith:
Closing this out, I think a `warn` is optimal at the moment. Testing frameworks are moving towards showing the warning by default so that users can catch deprecations early.

PR #61, opened 2018-10-01
various changes to path machinery

jcmgray:
## Description
In the process of comparing/benchmarking the new 'cheap' path (#60) I thought various consistency/convenience updates to the path machinery might be worthwhile. Some of these are bigger than others so definitely let me know what you think. 

1. Per #55, change the ``memory_limit`` default to unlimited. As I mentioned in the other thread, I think any time that the arrays are going to get too big for RAM, resorting to einsum is going to be so exponentially slow that a ``MemoryError`` might be preferable. We could add ``'warn'/'system'`` versions that take system memory into account, the only thing is that this requires checking the output dtype and getting the systems free memory which is a bit of overhead. This change does make 'optimal' a bit slower, but at the same time, it possibly wasn't obvious previously that the path found might not have been globally optimal. Happy to split/delay this into a later PR if necessary (@dgasmith?).

2. Deprecated (with a warning) the ``path=`` keyword argument in ``contract_path`` in favour of ``optimize=``. Other functions, including in ``numpy`` use ``optimize`` so I think this makes sense to have a matching signature between ``contract, contract_path``, & ``contract_expression``.

3. Factor the full path information into a class (``PathInfo``) and update it to allow printing flop costs ``>1e307`` (which previously errored). This allows access to the various path costs programatically (e.g. ``path.opt_cost``, ``path.largest_intermediate`` etc) and makes the default repr the full printed info so no need to ``print(path)``. This is a breaking change if anyone was using the ``path`` as a proper string (e.g. @fritzo I saw you were regexing this) but ultimately should negate the need to do that (and a cheap fix is just ``str(path)``).

4. Add helper function ``oe.helpers.rand_equation``. This is a just a private function that can be helpful to generate large random expressions with variable connectivity and is thus might be useful for testing.

5. Finally, the ``cupy`` tests weren't working - I think it needs to be imported explicitly, and before other GPU libs, in order to initialize CUDA properly, so I've done that.

## TODO:
- [x] Note the ``memory_limit`` choice in the docs somewhere?
- [x] Test the ``rand_equation`` generator / update some tests to use it?

## Status
- [x] Ready to go

### DIFF
--- a/opt_einsum/contract.py
+++ b/opt_einsum/contract.py
@@ -15,6 +15,68 @@ from . import sharing
 __all__ = ["contract_path", "contract", "format_const_einsum_str", "ContractExpression", "shape_only", "shape_only"]
 
 
+class PathInfo:
+
+    def __init__(self, contraction_list, input_subscripts, output_subscript,
+                 indices, scale_list, naive_cost, opt_cost, size_list):
+        self.contraction_list = contraction_list
+        self.input_subscripts = input_subscripts
+        self.output_subscript = output_subscript
+        self.indices = indices
+        self.scale_list = scale_list
+        self.naive_cost = naive_cost
+        self.opt_cost = opt_cost
+        self.largest_intermediate = max(size_list)
+
+    def __repr__(self):
+        from decimal import Decimal
+
+        # naive costs / speedups can easily reach >~ 1e308, need exact arithmetic
+        naive_cost = Decimal(self.naive_cost)
+        opt_cost = Decimal(self.opt_cost)
+        speedup = naive_cost / opt_cost
+        largest_intermediate = Decimal(self.largest_intermediate)
+
+        # Return the path along with a nice string representation
+        overall_contraction = self.input_subscripts + "->" + self.output_subscript
+        header = ("scaling", "BLAS", "current", "remaining")
+
+        path_print = "  Complete contraction:  {}\n".format(overall_contraction)
+        path_print += "         Naive scaling:  {}\n".format(len(self.indices))
+        path_print += "     Optimized scaling:  {}\n".format(max(self.scale_list))
+        path_print += "      Naive FLOP count:  {:.3e}\n".format(naive_cost)
+        path_print += "  Optimized FLOP count:  {:.3e}\n".format(opt_cost)
+        path_print += "   Theoretical speedup:  {:3.3f}\n".format(speedup)
+        path_print += "  Largest intermediate:  {:.3e} elements\n".format(largest_intermediate)
+        path_print += "-" * 80 + "\n"
+        path_print += "{:>6} {:>11} {:>22} {:>37}\n".format(*header)
+        path_print += "-" * 80
+
+        for n, contraction in enumerate(self.contraction_list):
+            inds, idx_rm, einsum_str, remaining, do_blas = contraction
+            remaining_str = ",".join(remaining) + "->" + self.output_subscript
+            path_run = (self.scale_list[n], do_blas, einsum_str, remaining_str)
+            path_print += "\n{:>4} {:>14} {:>22} {:>37}".format(*path_run)
+
+        return path_print
+
+
+def _choose_memory_arg(memory_limit, size_list):
+    if memory_limit == 'max_input':
+        return max(size_list)
+
+    if memory_limit is None:
+        return int(1e20)
+
+    if memory_limit < 1:
+        if memory_limit == -1:
+            return int(1e20)
+        else:
+            raise ValueError("Memory limit must be larger than 0, or -1")
+
+    return int(memory_limit)
+
+
 def contract_path(*operands, **kwargs):
     """
     Evaluates the lowest cost einsum-like contraction order.
@@ -25,7 +87,7 @@ def contract_path(*operands, **kwargs):
         Specifies the subscripts for summation.
     *operands : list of array_like
         These are the arrays for the operation.
-    path : bool or list, optional (default: ``auto``)
+    optimize : bool or list, optional (default: ``auto``)
         Choose the type of path.
 
         - if a list is given uses this as the path.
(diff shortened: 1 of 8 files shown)


opt_einsum/contract.py
```diff
@@ -15,6 +15,68 @@
...
+        path_print += "{:>6} {:>11} {:>22} {:>37}\n".format(*header)
+        path_print += "-" * 80
+
+        for n, contraction in enumerate(self.contraction_list):
+            inds, idx_rm, einsum_str, remaining, do_blas = contraction
+            remaining_str = ",".join(remaining) + "->" + self.output_subscript
+            path_run = (self.scale_list[n], do_blas, einsum_str, remaining_str)
+            path_print += "\n{:>4} {:>14} {:>22} {:>37}".format(*path_run)
+
+        return path_print
```
fritzo: nit: you could avoid quadratic growth by using `lines.append(...)` and finally `return '\n'.join(lines)`

opt_einsum/contract.py
```diff
@@ -15,6 +15,68 @@
 __all__ = ["contract_path", "contract", "format_const_einsum_str", "ContractExpression", "shape_only", "shape_only"]
 
 
+class PathInfo:
```
fritzo: nit: Python 2 convention is to always inherit from `object`:
```py
class PathInfo(object):
    ...
```

jcmgray:
@fritzo, great, I've made those couple of changes - thanks for pointing them out.

opt_einsum/contract.py
```diff
@@ -15,6 +15,70 @@
...
+
+        path_print = [
+            "  Complete contraction:  {}\n".format(overall_contraction),
+            "         Naive scaling:  {}\n".format(len(self.indices)),
+            "     Optimized scaling:  {}\n".format(max(self.scale_list)),
+            "      Naive FLOP count:  {:.3e}\n".format(naive_cost),
+            "  Optimized FLOP count:  {:.3e}\n".format(opt_cost),
+            "   Theoretical speedup:  {:3.3f}\n".format(speedup),
+            "  Largest intermediate:  {:.3e} elements\n".format(largest_intermediate),
+            "-" * 80 + "\n",
```
dgasmith: +1 for `format`. If you thinking about it can we replace the old `%` syntax elsewhere through the code so we can be consistent? Ancillary point, we can spin this off in its own issue.

opt_einsum/contract.py
```diff
@@ -111,12 +175,18 @@ def contract_path(*operands, **kwargs):
...
-    valid_contract_kwargs = ['path', 'memory_limit', 'einsum_call', 'use_blas']
+    valid_contract_kwargs = ['optimize', 'path', 'memory_limit', 'einsum_call', 'use_blas']
     unknown_kwargs = [k for (k, v) in kwargs.items() if k not in valid_contract_kwargs]
     if len(unknown_kwargs):
         raise TypeError("einsum_path: Did not understand the following kwargs: %s" % unknown_kwargs)
 
-    path_type = kwargs.pop('path', 'auto')
+    if 'path' in kwargs:
+        import warnings
+        warnings.warn("The 'path' keyword argument is deprecated in favor of 'optimize'.", DeprecationWarning)
```
dgasmith: Agree, this is a good change overall.

opt_einsum/contract.py
```diff
@@ -350,11 +391,14 @@ def contract(*operands, **kwargs):
...
           the number of terms in the contraction.
 
-    memory_limit : int or None (default : None)
-        The upper limit of the size of tensor created, by default, this will be
+    memory_limit : {None, int, 'max_input'} (default: None)
         Give the upper bound of the largest intermediate tensor contract will build.
-        By default (None) will size the ``memory_limit`` as the largest input tensor.
-        Users can also specify ``-1`` to allow arbitrarily large tensors to be built.
+
+        - if None or -1, there is no limit.
```
dgasmith: Can we change the text style here to follow more like:

```
- ‘None’ or -1 means there is no limit
- ‘max_input’ means the limit is set as the size of the largest input tensor
...
The default is `None`.
```

opt_einsum/helpers.py
```diff
@@ -167,3 +168,95 @@ def flop_count(idx_contraction, inner, num_terms, size_dictionary):
...
+    Returns
+    -------
+    eq : str
+        The equation string.
+    shapes : tuple[tuple[int]]
+        The array shapes.
+
+    Examples
+    --------
+    >>> eq, shapes = rand_equation(n=10, reg=4, n_outer=5, seed=42)
```
dgasmith: Should be `rand_equation(10, 4, 5, seed=42)` I think?

opt_einsum/helpers.py
```diff
@@ -167,3 +168,95 @@ def flop_count(idx_contraction, inner, num_terms, size_dictionary):
...
+    """Generate a random contraction and shapes.
+
+    Parameters
+    ----------
+    n : int
+        Number of array arguments.
+    reg : int
+        Average connectivity of graph.
+    n_outer : int
+        Number of outer indices.
```
dgasmith: Can you expend on outer indices?

opt_einsum/helpers.py
```diff
@@ -167,3 +168,95 @@ def flop_count(idx_contraction, inner, num_terms, size_dictionary):
...
+
+def rand_equation(n, reg, n_outer, dmin=2, dmax=9, seed=None):
+    """Generate a random contraction and shapes.
+
+    Parameters
+    ----------
+    n : int
+        Number of array arguments.
+    reg : int
+        Average connectivity of graph.
```
dgasmith: Not sure what `reg` is meant to symbolize here, perhaps write this word out?

opt_einsum/helpers.py
```diff
@@ -167,3 +168,95 @@ def flop_count(idx_contraction, inner, num_terms, size_dictionary):
         op_factor += 1
 
     return overall_size * op_factor
+
+
+def rand_equation(n, reg, n_outer, dmin=2, dmax=9, seed=None):
```
dgasmith: There might be a slight mismatch between styles here `dmin` `nouter` vs `d_min`, `n_outer`. I think your current nomenclature is the lesser of the evils, but worth thinking about.

dgasmith:
It might be good to write a few tests that use the random path tech with a static seed and compares that the results are the same across all paths.

opt_einsum/helpers.py
jcmgray: This works fine for me leaving them all as keyword arguments?

opt_einsum/tests/test_contract.py
```diff
@@ -231,3 +229,17 @@ def test_contract_expression_with_constants(string, constants):
...
     out = expr(*ctrc_args)
     assert np.allclose(expected, out)
+
+
+@pytest.mark.parametrize("optimize", ['greedy', 'optimal'])
+@pytest.mark.parametrize("n", [3, 5])
+@pytest.mark.parametrize("reg", [3, 4])
+@pytest.mark.parametrize("n_out", [0, 2, 4])
+def test_rand_equation(optimize, n, reg, n_out):
+    eq, shapes = helpers.rand_equation(n, reg, n_out, d_min=2, d_max=5)
```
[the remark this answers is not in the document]
jcmgray: Thanks for spotting I missed this - fixed now.

jcmgray:
Great I'll merge.

Issue #67, opened 2018-10-17
Support more subscript symbols


liwt31:
The project now supports more than 52 subscripts, but only with integers, and it can quickly get confusing using so many integers. My suggestion is to support this kind of syntax:
```
arr1 = np.zeros((2, 3))
sub1 = '[left edge][bond]'
arr2 = np.zeros((3, 2))
sub2 = '[bond][right edge]'
out = '[left edge][right edge]'
opt_einsum.contract(arr1, sub1, arr2, sub2, out)
```

I know that it's equivalent with:
```
arr1 = np.zeros((2, 3))
sub1 = [0, 1]
arr2 = np.zeros((3, 2))
sub2 = [1, 2]
out = [0, 2]
```

However, IMHO, when doing contraction with lots of tensor which have different meanings, it'll be nice to give them a proper name.

I'm more than happy to prepare a PR for this, I think it's not extemely difficult.

jcmgray:
Yes that would be how to specify the output indices - for other contraction syntaxes this doesn't need to specified, its just the alphabetic order of any uncontracted (single) indices, I imagine alphabetic- or input-order would be a good default here, as well.

> I am deeply impressed by the quimb.tensor module. Have you ever thought about separating it as an individual package?

Thanks! Yes it occurred to me but I haven't had the time or need to decouple it quite yet..

dgasmith:
@jcmgray Should we consider moving the edge/bond syntax into `opt_einsum` or leave it elsewhere? I would be a bit concerned that a third input form on `contact` could become unwieldy as we have the above mentioned integer input as well.

jcmgray:
Yes I am in two minds too a bit. One thing that is possible is just to broaden the current case to anything hashable (which includes ``ints``) so it would not be two separate input forms. Here is useful snippet that could replace the current logic:

```python
>>> symbols = (oe.get_symbol(i) for i in itertools.count())
>>> symbol_map= defaultdict(lambda: next(symbols))
>>> symbol_map['left'] 
'a'
>>> symbol_map['right]
'b'
>>> symbol_map['left']
'a'
```
then you could replace the line
```python
subscripts += get_symbol(s)
```
just with
```python
subscripts += symbol_map[s]
```
at basically no cost.

dgasmith:
Closed by #68.

PR #68, opened 2018-10-18
Support more symbols for subscripts (#67)

liwt31:
* Add more symbol support based on hash

* Refactor the converting to subscripts routine a little bit

* Update tests

## Description
Make opt_einsum support anything hashable as subscripts, such as:
```python
contract(arr1, ['left edge', 'bond'], arr2, ['bond', 'right edge'])
```
See #67 

## Questions
- [x] I've tried to add the new feature to the document, but it seems the document does not have a chapter describing the inputs. Maybe I've left out something or maybe this should be solved in a different PR?

## Status
- [x] Ready to go

### DIFF
--- a/opt_einsum/parser.py
+++ b/opt_einsum/parser.py
@@ -189,7 +189,7 @@ def possibly_convert_to_numpy(x):
 
 
 def convert_subscripts(old_sub, symbol_map):
-    """Convert user custom subscripts list to subscript string according to `symbol_map`. Raises TypeError if any subscripts is not hashable.
+    """Convert user custom subscripts list to subscript string according to `symbol_map`.
 
     Examples
     --------
@@ -203,10 +203,8 @@ def convert_subscripts(old_sub, symbol_map):
         if s is Ellipsis:
             new_sub += "..."
         else:
-            try:
-                new_sub += symbol_map[s]
-            except TypeError:
-                raise TypeError("For this input type lists must contain either hashable object or Ellipsis")
+            # no need to try/except here because symbol_map has already been checked
+            new_sub += symbol_map[s]
     return new_sub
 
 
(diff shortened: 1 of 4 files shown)


jcmgray:
@dgasmith I think this feels like a natural extension/inclusion to me. 

I did think of an extra detail that might complicate things a bit - currently the ``int`` order specifies the output order, so:
```python
np.einsum(a, [1, 0])
```
performs the transpose and ``opt_einsum`` should match this behaviour. The current implementation will map this to ``'ab'`` which performs no transpose. Potential solutions:

1. Use ``get_symbol(s)`` if ``int`` supplied and allow no mixing with other hashables
2. Compute the output symbols explicitly if not specified, I think we might do this elsewhere already
3. Somehow sort the ``symbol_map`` - not clear how to do this for mixed hashable types

dgasmith:
I agree that this turned out to be really quite elegant to integrate and should not cause much confusion. For the above issue, @liwt31 can you add a test like `opt_einsum.contract(np.zeros((2, 3)), [1, 0]).shape == (3, 2)` to make sure this always works? Might be good to add a test that uses `["b", "a"]` as our indices as well.

On the input issue would you mind starting a document that describes inputs? We can handle the canonical cases, if you just document this case it would be a great starting point.

@jcmgray Is allowing mixed hashable types really a flexibility that we need to handle?

jcmgray:
> @jcmgray Is allowing mixed hashable types really a flexibility that we need to handle?

No no, I think 99% use cases are just going to be strings. It would just have been nice and simple to essentially see integers with everything else as the same things: hashable.

So we just need to efficiently identify the integer case. Would just dispatching on the very first index be acceptable?:
```python
if isinstance(subscript_list[0][0], numbers.Integral):
    symbol_map = defaultdict(get_symbol)
else:
    symbols = (get_symbol(i) for i in itertools.count())
    symbol_map = defaultdict(lambda: next(symbols))
```
Or do we want the extra overhead of checking all the arguments are ``int`` or not?

For the docs, it will need to mention that the default order will be based on the input order rather than sorted order if non-integers are used.

dgasmith:
I like it, and seems fairly readable with a short explanation.

jcmgray:
> I suppose we all prefer 1?

I think so. No use cases really come to mind where this would be critical, so best to be more stringent I guess. And it if does come up later, would be very easy to add.

dgasmith:
This pull request **introduces 1 alert** when merging 33a09d6b2e23783a8ec1d0cb597f93dea7293c5f into 02d15b7fee68decec69950f9b04de993900d87bd - [view on LGTM.com](https://lgtm.com/projects/g/dgasmith/opt_einsum/rev/pr-3fc36f0aa0bd59bf0617ea83cc9f1d50521c8471)

**new alerts:**

* 1 for Unused import

---

*Comment posted by [LGTM.com](https://lgtm.com)*

dgasmith:
I agree 1) sounds good for now.

opt_einsum/parser.py
```diff
@@ -230,28 +251,20 @@ def parse_einsum_input(operands):
...
-                if s is Ellipsis:
-                    subscripts += "..."
-                elif isinstance(s, int):
-                    subscripts += get_symbol(s)
-                else:
-                    raise TypeError("For this input type lists must contain either int or Ellipsis")
-            if num != last:
-                subscripts += ","
-
+        try:
```
dgasmith: Can you break up these try/except and if blocks with newlines between them with a small comment describing why the steps are needed?

liwt31: Sure, I was going to

liwt31:
> On the input issue would you mind starting a document that describes inputs? We can handle the canonical cases, if you just document this case it would be a great starting point.

@dgasmith I started some document on the input format. The canonical cases are now left blank.

liwt31:
I think that's all. Thank you for your comments and advice!

jcmgray:
Looks good to me as well, thanks for this!

PR #72, opened 2018-10-20
input fixes and docs

### LINKED ISSUE #70: Finish out input documentation
Input documentation was started in #68. However, the canonical `einsum` string and integer-based contractions specifications are not yet documented. The current stub can be found in `docs/source/input_format.rst`.

### LINKED ISSUE #71: let contract_expression use interleaved input
Currently, ``contract_expression`` only accepts the 'equation' style input. This is an easy fix that I will do shortly. As part of that I can flesh out the docs on input (#70), unless you are already planning that @dgasmith ?

jcmgray:
## Description
Makes ``contract_expression`` accept interleaved-style input (fixes #71) and adds general documentation on the input formats (resolves #70) following on from #67.

## Questions
- [ ] I've moved the input section into the main doc sections as this seemed quite natural?

## Status
- [x] Ready to go

### DIFF
--- a/docs/source/input_format.rst
+++ b/docs/source/input_format.rst
@@ -24,8 +24,8 @@ followed by the array arguments:
     >>> z.shape
     (5, 2)
 
-The main difference is that thousands of symbols in the form of unicode
-characters can be used:
+However, in addition to the standard alphabet, ``opt_einsum`` also supports
+unicode characters:
 
 .. code:: python
 
@@ -33,7 +33,8 @@ characters can be used:
     >>> oe.contract(eq, x, y).shape
     (5, 2)
 
-One way to access these programmatically is through the function
+This enables access to thousands of possible index labels. One way to access
+these programmatically is through the function
 :func:`~opt_einsum.parser.get_symbol`:
 
     >>> oe.get_symbol(805)
--- a/docs/source/input_format.rst
+++ b/docs/source/input_format.rst
@@ -2,13 +2,71 @@
 Input Format
 ============
 
-The ``opt_einsum`` package is a drop-in replacement for the ``np.einsum`` function
-and supports all input formats that ``np.einsum`` supports.
+The ``opt_einsum`` package is a drop-in replacement for the ``np.einsum``
+function and supports all input formats that ``np.einsum`` supports. There are
+two styles of input accepted, a basic introduction to which can be found in the
+documentation for :func:`numpy.einsum`. In addition to this, ``opt_einsum``
+extends the allowed index labels to unicode or arbitrary hashable, comparable
+objects in order to handle large contractions with many indices.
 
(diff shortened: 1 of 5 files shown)


dgasmith:
This pull request **introduces 1 alert** when merging aa47e5b549d7ec464bf975db3815f0444f84caf4 into 7daf290b236d3f464845152b25ee58509410d976 - [view on LGTM.com](https://lgtm.com/projects/g/dgasmith/opt_einsum/rev/pr-eed860d86e7d7251d2af7be5199ad3c1bf274446)

**new alerts:**

* 1 for Syntax error

---

*Comment posted by [LGTM.com](https://lgtm.com)*

docs/source/input_format.rst
```diff
@@ -2,13 +2,71 @@
...
+
+.. code:: python
+
+    >>> eq = 'ijk,jkl->li'
+    >>> x, y = np.random.rand(2, 3, 4), np.random.rand(3, 4, 5)
+    >>> z = oe.contract(eq, x, y)
+    >>> z.shape
+    (5, 2)
+
+The main difference is that thousands of symbols in the form of unicode
```
dgasmith: Maybe explicitly say "In addition to standard characters, `opt_einsum` also supports unicode..." or the like?

opt_einsum/contract.py
```diff
@@ -805,6 +805,9 @@ def contract_expression(subscripts, *shapes, **kwargs):
             raise ValueError("'{}' should only be specified when calling a "
                              "`ContractExpression`, not when building it.".format(arg))
 
+    if not isinstance(subscripts, compat.strings):
```
dgasmith: Any reason why we just do not call `parse_einsum_input` here? It would be good, if possible to always parse the input at a single location. I think current `parse_einsum_input` does not yet support non-array like inputs such as tuples, but this should easy to correct I believe.

docs/source/input_format.rst
jcmgray: Yes good idea.

opt_einsum/contract.py
jcmgray: So, it's just because ``contract_expression`` routes through ``contract`` and all its machinery once the shapes and constants are processed and thus ``parse_einsum_input`` will be called again anyway. The only reason ``convert_interleaved_input`` is needed here (which will then not be called again), is to simplify the ``constants`` processing.

One option might be have ``contract`` accept a tuple of the parsed inputs, or a ``is_parsed`` option or something, so that ``contract_expression`` could do the parsing and call ``contract`` with the guarantee that it is already parsed.

dgasmith: It might be good to break out the inner contract loops at some point to make the actual iterations more flexible.

Issue #64, opened 2018-10-02
Allow custom cost objective functions


jcmgray:
Experimenting with #60 and #63 made me think that allowing custom cost objective functions would be neat. The current the cost function greedy uses is
```python
size12 - size1 - size2
```
though as @fritzo notes, just using
```python
size12
```
made some contractions more deterministic and thus easier to cache.

Generally you might have:
```python
A * size12 - B * size1 - C * size2
```
with e.g. ``A``, ``B``, ``C`` random modifiers for #63. One might even try and learn the values using the contraction complexity / flop cost as a loss function.

dgasmith:
Looking at it I think you are right, we could implement these as `if` statements and hardcode them in, but ultimately passing functions in would likely be more performant and easier to modify. We could have a stable of functions as well that could be shared between algorithms:

```
objective_functions = {
    "canonical": lambda s12, s1, s2: s12 - s1 - s2,
    "size12": lambda s12, s1, s1: s12,
    ....
}

**Comment by fritzo** [comment]
Hmm the objective functions I've been experimenting with don't fit into this framework, and might be difficult to abstract out without a very wide interface. E.g. I've been playing with objective function terms including:
- `len(k12 & output)`, encouraging contractions that result in few dimensions
- `len(k12 - dim_ref_counts[2])`, encouraging contractions on the periphery of the connectivity graph
- `len(k12) - len(k1 | k2)`, i.e. dim pressure, analogous to memory pressure
- "run Kruskal's algorithm; encourage contractions farthest away from contraction dims"

**Comment by jcmgray** [comment]
Probably all the information that might be needed is already passed to ``_get_candidate``? At least for the moment. In terms of performance, adding something like:
```python
if cost_func is None:
    cost = sizes[k12] - sizes[k1] - sizes[k2]
else:
    cost = cost_func(sizes, k12, k1, k2, ...)
```
should not add much overhead in the default case.

Issue #63, opened 2018-10-02
Use best of many random greedy paths


jcmgray:
In quantum computing simulations particularly, finding the optimal contraction sequence has a huge effect on what is computable - the amount of memory and time required is exponential in how 'bad' the path found is. The greedy approach is generally worse than the optimal approach but it is cheap to compute (and even cheaper with #60!).

If we introduce a random greedy algorithm by tweaking the cost objective function, we get a variety of different paths with varying costs and we can just pick the best one. I had a little play with contracting a 49 qubit depth 20 circuit (661 tensors overall) and randomly adjusting each cost by +- 5%.

The following are 'contraction complexities' (maximum rank of tensor produced - ``2^this`` is the memory required which is also a rough proxy for computation time) on the left and frequency on the right from 100 runs.
```
[(25, 1),
 (27, 1),
 (28, 3),
 (29, 6),
 (30, 10),
 (31, 16),
 (32, 6),  # <- this is what the current greedy finds
 (33, 9),
 (34, 21),
 (35, 8),
 (36, 8),
 (37, 1),
 (38, 4),
 (39, 2),
 (40, 1),
 (42, 3)]
```
So the best contraction found vs the current contraction found requires ~128x less memory and computation, and turns the simulation from 'super'-computable to 'laptop'-computable.

dgasmith:
This is found by modifying A/B/C as shown in #64 or is the randomization something else?

jcmgray:
This was done by setting the cost as:
```python
cost = (
    (1 + 0.05 * randn()) * size12 - 
    (1 + 0.05 * randn()) * size1 - 
    (1 + 0.05 * randn()) * size2
)
```
You seem to get similar results for small amounts of randomness, then it gets much worse for large amounts of randomness (>50%).

jcmgray:
It would also be a very simple way of parallelizing the path finding aspect of ``opt_einsum``.

dgasmith:
That is pretty impressive, but in retrospect not terrible surprising I guess as the so called "shake" category of algorithms are based along similar ideas. We can change our existing algorithms to include this fairly easily and have a "shake" algorithm that would perform the meta evaluation of multiple runs. `shake-greedy`, `shake-cheap` etc.

If you have time, could you prototype out an example? I also have a bit this weekend to hack out some code as this look fairly straightforward.

jcmgray:
Happy to implement this quite soon - I think it might be good to have #60 and #64 in before.

[...]

PR #73, opened 2018-10-21
[WIP] recursive, depth-first optimizer

jcmgray:
## Use recursive, depth-first optimizer for optimal & possibly 'greedy2'

I had a play around with making the ``opt_einsum`` optimal path finder a bit more efficient. The basic idea is:

1. Use a recursive search to simplify the logic and find paths 'depth-first', i.e. follow a path all the way to its termination before searching for the next one.
2. Allowed by this, terminate searches as soon as the path cost increases above the best found so far, sieving many known-bad paths

The results are pretty promising, current optimal:

```python
>> eq, shapes = oe.helpers.rand_equation(8, 3, seed=42)
>> views = [np.random.rand(*s) for s in shapes]
>> %time path = oe.contract_path(eq, *views, optimize='optimal')[1]
CPU times: user 25.3 s, sys: 716 ms, total: 26 s
Wall time: 26 s
>>> path.opt_cost
130316
```

and with this recursive strategy:

```python
>> %time path = oe.contract_path(eq, *views, optimize='roptimal')[1]
CPU times: user 107 ms, sys: 324 µs, total: 107 ms
Wall time: 108 ms
>>> path.opt_cost
130316
```

The fundamental scaling is still factorial, but it should allow optimal-paths to be realistically found for expressions of up to 10 tensors. Possibly other heuristics can be used to sieve paths eagerly as well.

One additional option this opens up is to recurse only into the ``n`` best possible contractions (current greedy/eager being ``n=1``). Thus you could have a ``'greedy2'``, ``'greedy3'`` etc, which scale like ``2**n``, ``3**n`` which while exponential (modulo sieving), is significantly more manageable than factorial scaling. These might use custom cost functions #64 and random shaking as well #63.

Anyway, thought I'd share this WIP initial implementation. My suggestion might be to have this approach eventually replace optimal, with ``'greedy'``, ``'greedy2'``... using it also.

cc @fritzo --- took some inspiration from your ``'eager'`` path if you have any input!

TODO:
---------
- [x] apply some caching?
- [ ] keep track of ``keep`` indices rather than computing each time?
- [x] handle memory_limit case properly? (i.e. contract all remaining tensors at once)


dgasmith:
This is pretty cool and a good way to trim a lot of bad paths early. You are right that it provides a lot of interesting paths to explore in terms of different cost functions and path randomness.

Something to try would be to evaluate a greedy path at the very start and use that as a sieving upper bound.

opt_einsum/paths.py
```diff
@@ -88,6 +89,62 @@ def optimal(input_sets, output_set, idx_dict, memory_limit):
...
+                best['path'] = path
+            return
+
+        # check all possible remaining paths
+        for i, j in itertools.combinations(remaining, 2):
+            k12, flops12 = _calc_k12_flops(inputs, output, remaining, i, j, size_dict)
+
+            # sieve based on current best flops
+            new_flops = flops + flops12
+            if new_flops > best['flops']:
```
fritzo: you could broaden to `if new_flops >= best['flops']`

opt_einsum/paths.py
```diff
@@ -88,6 +89,62 @@ def optimal(input_sets, output_set, idx_dict, memory_limit):
...
+                continue
+
+            # sieve based on memory limit
+            if memory_limit is not math.inf:
```
fritzo: note `math.inf` is not defined in Python 2. Also you shouldn't use `is` to compare floats. A safer version is:
```py
if memory_limit < float('inf'):
```

opt_einsum/paths.py
```diff
@@ -88,6 +89,62 @@ def optimal(input_sets, output_set, idx_dict, memory_limit):
...
 
 
+def _calc_k12_flops(inputs, output, remaining, i, j, size_dict):
+    k1, k2 = inputs[i], inputs[j]
+    either = k1 | k2
+    two = k1 & k2
+    one = either - two
+    keep = frozenset.union(output, *(inputs[p] for p in remaining if p not in (i, j)))
+
+    k12 = one | (two & keep)
```
fritzo: Hmm can you help me understand this? I would think
```
k12 == either & keep == (one | two) & keep
```
whereas `one | (two & keep)` could accidentally preserve a dim that should not be kept 😕

opt_einsum/paths.py
jcmgray: Ah thanks, my python2 knowledge is quite lacking...

opt_einsum/paths.py
jcmgray: Yes I up ending confusing myself when writing this (then it worked for my contractions so I forgot about it!) - glad you had a look over. I was coming from the standard tensor notation where in general ``k12 = one``, then I added in ``two & keep`` to take care of triple+ indices, forgetting that ``output`` is already processed and in ``keep``. I'll change to:
```python
k12 = either & keep
```
Some kind of ``dim_ref_count`` as you had in ``'eager'`` might be nice here, the trick is just that each of  the factorially many paths would need its own - a bit cumbersome.

jcmgray:
> Something to try would be to evaluate a greedy path at the very start and use that as a sieving upper bound.

Yes definitely, so if you sort the contractions/use a heap at each step (as you might before recursing into only the best ``N``) , the very first path that is analysed will be the greedy path, setting a reasonable sieve.

--- 

Regarding caching, am I right in thinking that for a given ``k1``, ``k2`` pair, regardless of what other contractions have happened, the resulting ``k12`` and flop cost will be the same? This would be a nice saving. It's definitely true for 'standard' contractions where indices appear no more than twice, and think so otherwise but not totally sure, (@dgasmith / @fritzo)?

dgasmith:
`abc, abc -> abc` vs `abc, abc ->`. The cost of the latter requires an extra factor of two (unless you use FMA which you probably do...) as you need an extra addition per element.

jcmgray:
> abc, abc -> abc vs abc, abc ->. The cost of the latter requires an extra factor of two (unless you use FMA which you probably do...) as you need an extra addition per element.

Aha while you are right here, these two will actually never appear within the same contraction as the output is constant throughout. If indices should be kept as they appear on other inputs

So I've tentatively enabled caching and it seems to produce the correct results so far (and reduces the time for the path finding above to ``~25ms``.

dgasmith:
With caching we might need to watch out how much we are caching here. The factorial compute is associated with factorial memory requirements. The speedups are quite good however with caching!

jcmgray:
I've updated the optimal approach a bit and implemented a recursive greedy strategy (orientated around greedily taking *multiple* paths at once rather than just a single one).

Some broadlly illustrative numbers using the setup above:

## new optimal 

(can realistically handle up to about ~10 tensors):
```python
>>> oe.contract_path(eq, *views, optimize='roptimal')[1].opt_cost
130316

>>> %timeit oe.contract_path(eq, *views, optimize='roptimal')[1]
15.6 ms ± 273 µs per loop (mean ± std. dev. of 7 runs, 100 loops each)
```

## 'greedy-all' / 'branch-all'

Quite similar to optimal but only assess inner products, and actively terminate paths if they relatively become bad rather than waiting (can realistically handle up to about ~14 tensors):

```python
>>> oe.contract_path(eq, *views, optimize='rgreedy')[1].opt_cost
130362

>>> %timeit oe.contract_path(eq, *views, optimize='rgreedy')[1]
1.26 ms ± 17.2 µs per loop (mean ± std. dev. of 7 runs, 1000 loops each)
```

## 'greedy2' / 'branch-2'

As above, but only keep the two best contractions upon each recurse (can realistically handle up to about ~20 tensors):

```python
>>> oe.contract_path(eq, *views, optimize='rgreedy2')[1].opt_cost
130362

>>> %timeit oe.contract_path(eq, *views, optimize='rgreedy2')[1]
757 µs ± 3.45 µs per loop (mean ± std. dev. of 7 runs, 1000 loops each)
```

And then the old greedy and eager for comparison:

```python
>>> # greedy
>>> oe.contract_path(eq, *views, optimize='greedy')[1].opt_cost
520152

>>> %timeit oe.contract_path(eq, *views, optimize='greedy')[1]
423 µs ± 4.13 µs per loop (mean ± std. dev. of 7 runs, 1000 loops each)

>>> # eager
>>> oe.contract_path(eq, *views, optimize='eager')[1].opt_cost
815064

>>> %timeit oe.contract_path(eq, *views, optimize='eager')[1]
429 µs ± 3.08 µs per loop (mean ± std. dev. of 7 runs, 1000 loops each)
```

So the new recursive greedy (which maybe needs a better name to differentiate? ``'branch-all'``, ``branch-2``?) works quite well as an intermediate point between quality and run-time.

I imagine now:
1. Just replacing the ``optimal`` path, the new one I think finds the exact same paths.
2. Adding ``'branch-all'``, ``'branch-2'`` as options.
3. Removing the old ``greedy`` implementation and making it point to ``eager`` which is faster while finding roughly the same quality of contractions
4. Tweaking ``'auto'`` to dispatch to optimal, then the branching/recursive greedy, then eager.

opt_einsum/paths.py
```diff
@@ -14,6 +14,9 @@
 __all__ = ["optimal", "greedy", "eager"]
 
 
+inf = float('inf')
```
dgasmith: I dont think we declare this enough times it is worth declaring a global.

opt_einsum/paths.py
```diff
@@ -88,6 +91,165 @@ def optimal(input_sets, output_set, idx_dict, memory_limit):
     return path
 
 
+def _calc_k12_flops(inputs, output, remaining, i, j, size_dict):
```
dgasmith: Can you do a mini doc on here and the below similar to others?

opt_einsum/paths.py
```diff
@@ -88,6 +91,165 @@ def optimal(input_sets, output_set, idx_dict, memory_limit):
...
+    def _greedy_iterate(path, inputs, remaining, flops):
+
+        # reached end of path (only ever get here if flops is best found so far)
+        if len(remaining) == 1:
+            best['flops'] = flops
+            best['path'] = path
+
+        def _assess_candidate(k1, k2, i, j):
+            # find resulting indices and cost
+            try:
```
dgasmith: Can we do a `if (k1, k2) in result_cache` here and below? Although I guess it might be slower... maybe:

```
k12, flops12 = result_cache.setdefault((k1, k2), _calc_k12_flops(inputs, output, remaining, i, j, size_dict))
```

jcmgray: This is, marginally, the quickest I've found! I think because the key is more often in the cache than not - otherwise checking membership first would edge it in terms of performance. The ``setdefault`` is definitely elegant but I think it has to evaluate the ``default`` arg everytime, even if ``k1, k2`` is in the cache.

dgasmith:
Can you check to make sure the new and old optimal all give the same results by trying maybe 1-10k random paths for 1-6 tensors? If that passes I think we should actually look at removing the current optimal path, I believe this does exactly the same thing and will have much better performance characteristics due to more aggressive pruning.

I like the `branch` name for the new category of optimizers as I don't quite think the characteristics are similar to greedy.

It doesn't hurt to have many different paths, but the rate we are growing is pretty steep and it is not clear if the various paths are "better" than another and what "better" really means. We will have to sum this up to users in a succinct way at some point and make recommendations. Work is extremely busy at the moment so I haven't had a chance to dive into this.

opt_einsum/paths.py
dgasmith: Maybe `value = cache.get(key) or cache.setdefault(key, func(...))`?

jcmgray: I think my comment above applies still, the ``try...except`` with standard ``__getitem__``/``__setitem__`` syntax performs best overall. The way that python works means ``func(...)`` will be eagerly called and supplied to ``setdefault`` even if it is not needed sadly.

jcmgray:
> Can you check to make sure the new and old optimal all give the same results by trying maybe 1-10k random paths for 1-6 tensors?

I've now done this for various contractions from ``rand_equation`` and the flop cost found is always identical (though occasionally a different path is found - an 'equally optimal' one). I might add 3+ index contractions to be very certain at some point later.

opt_einsum/paths.py
jcmgray: From running a sample ``optimal`` the timings I get are:
```
try-exc / setdefault : 127 ms ± 2.26 ms
try-exc /   setitem  : 131 ms ± 4.18 ms
 not-in /   setitem  : 141 ms ± 1.33 ms
 not-in / setdefault : 148 ms ± 3.86 ms
  get   /   setitem  : 146 ms ± 1.97 ms
  get   / setdefault : 149 ms ± 2.83 ms
```
so there does seem to be a 5-10% advantage - I think just because the cache is hit more often than missed. However, these timings are marginal (maybe machine dependant etc), and though I have always thought of ``try/except`` as 'pythonic', completely happy to defer to whichever you want as long as you don't mind a possible slight performance hit!

dgasmith: Interesting do you precompute the key in your timings `key = str((k1, k2))`?

Either way, can you tuck this functionality into a util function so that we can revisit later?

fritzo: In my experience, the two standard patterns are: `try...except` if keys are usually present, and `if ... in ... else` if keys are usually absent. `setdefault` is only used if value computation is free.

jcmgray: > Interesting do you precompute the key in your timings key = str((k1, k2))?

I'm using ``key = (k1, k2)`` now (marginal saving over forming the tuple twice), this is already hashable due to being frozensets so applying ``str`` or ``hash`` doesn't help I find. 

> Kind of interesting that you get this setdefault/setitem reversal in your timings.

I suspect this is possibly just noise? I'd be happy to use either but I fell like the standard setitem is simpler here as we know the key is not in the cache already.

> Either way, can you tuck this functionality into a util function so that we can revisit later?

So, I did try this with a ``lazy_setdefault`` function, but you get quite a slow down - these few lines are a real hot spot where these micro optimizations make a big difference it seems!

Overall I'm leaning towards leaving it roughly as it is?

```python
key = (inputs[i], inputs[j])
try:
    k12, flops12 = result_cache[key]
except KeyError:
    k12, flops12 = result_cache[key] = _calc_k12_flops(inputs, output, remaining, i, j, size_dict)
```

dgasmith: Alright, at some point I am just complaining for no good reason. If you could cast it in a function that did the try/except/lazy evaluate that would be best, but no reason to eat the overhead if it costs too much.

jcmgray:
Ok sorry for the delay I've updated this PR a bit:

1. made the recursive strategy the default optimal algorithm
2. made ``'greedy'`` point to the eager algorithm
3. added ``'branch-all'`` and ``'branch-2'`` as options
4. ``'auto'`` mode now uses :
    -  ``'optimal'`` for <= 4 inputs
    -  ``'branch-all'`` for <= 6 inputs
    -  ``'branch-2'`` for <= 8 inputs
    - ``'greedy'`` / ``'eager'`` otherwise.

   These cutoffs are chosen such that the path finding is always well below ``1ms``, but with nice increases in quality of contraction for up to 8 tensors.

Currently I've made ``'greedy'`` with a ``memory_limit`` set point to ``branch(..., nbranch=1)``, which does essentially the same thing but quite a lot less efficiently. The options here are:
* a) leave it like this
* b) just ignore ``memory_limit`` for greedy contractions (maybe with a warning if set?)
* c) modify ``eager`` to accept a ``memory_limit`` (most work and might make ``eager`` slower)

## TODO:
- [ ] Decide on a ``memory_limit`` strategy for greedy/eager
- [ ] Document the changes
- [ ] Make ``branch`` more customizable? I would envisage something like 

```python
contract(..., optimize=oe.BranchOptimizer(nbranch=3, flops_cutoff=1.5))
```
with a similar syntax maybe for #63. I suspect ``branch`` will mostly be used as part of the ``auto`` strategy so this might not be that important an addition.

jcmgray:
> @jcmgray I am unfortunately swamped at the moment and will not be able to get to this until the weekend.

No worries, no rush at all from my end.

> Cursory thoughts says we do need some sort memory_limit for greedy if possible, as that feature is supported in all others paths. Can you give me a quick rundown of why memory_limit and branch are not friends?

So, it's no particularly bad interaction with ``memory_limit`` and ``branch``, just that ``branch`` is orientated around exploring many paths at once, and when it is only doing a single path, ``eager``/``greedy`` has a much more efficient design that scales better. I'll might post a graph of some timings in a second.

At the moment, ``greedy`` with a ``memory_limit`` will work as expected, just not as quickly for big contractions - for me this is an OK compromise as we now have by default no ``memory_limit`` so it's more of a niche scenario (and a greedy approach can't explore the space of possibly lower memory paths as well anyway so it makes less sense).

> My thought is to get this in and prepare a review of the various branches so that we can show statistically that our selections are good after optimal. Once we have ascertained this we should be primed for another release.

Sounds good!

jcmgray:
I ran some quick timings (vertical lines denote proposed ``auto`` cutoffs from ``optimal -> branch-all -> branch-2 -> greedy/eager``):

opt_einsum_timings

With reference to points above note how ``branch-1`` scales a fair bit worse than ``greedy/eager``.

And the contraction quality increase (more is better):

opt_einsum_flops

NB. the low speedups are a bit misleading as (1) random graphs seem to be quite 'easy' to find paths for; and (2) it's really the worst-case speedup that is important rather than the mean speedup. However this graph does show the general trend that the various methods all find better contractions than a simple greedy approach  which I guess is reassuring.

As an aside, occasionally ``optimal`` does worse than ``greedy``? Slight inconsistencies in flop-counting maybe?

dgasmith:
That is a really great graph, mind if I tweet it out to get more opt_einsum interest?

While the graph is good, could you change it to a `pandas.describe()` or similar so that we can see min/max and percentiles? I wouldn't mind trying running some of these as well to find look into it a bit deeper. This is basically the kind of analysis that I was commenting on so thank you for already doing it!

It would be good to find the cases where it thinks `greedy` is better than `optimal` which should not be possible. I think you are right it could be incorrect flop clouting, though I thought we had found all of those.

dgasmith:
The outer product is actually memory bound and still quite small so raw FLOP count doesn't really kick in while the GEMM is compute bound, optimization via vendor BLAS, and likely uses multiple threads.

The short answer is it isn't very feasible to get a better handle on *time* rather than simple FLOP count without a lot of overhead. For example does the code take advantage of cache locality, SIMD instructions, FMA, advantage of superscalar architecture, etc. A good example is NumPy vs [HPTT](https://github.com/springer13/hptt), you can get 10-20x transpose speedup over NumPy. There are also BLAS libraries which do not require in-memory transposes and perform transpose in cache after being pulled from DRAM, see [TBLIS](https://github.com/devinamatthews/tblis).

For just NumPy our GEMM's would need to understand which vendor BLAS is used (MKL will start scaling like N^2.8 after a given point rather than N^3) or are they small enough to fall back to a simple loop implementation, which transposes do we need to perform to correctly execute the BLAS to begin with? Are we using einsum which uses explicit SSE2 vectorization while normal NumPy expressions might use AVX these days, Intel NumPy functions are also vectorized from SSE to AVX-512.

Not impossible, but this is some of the starting issues to begin to contend with. Would be great fun if I had some extra time on my hands...

jcmgray:
Yes definitely all that stuff seems too complex to assess (for me or in general!). 

Here's a slightly more specific question I guess -- those two contractions have a similar naive implementation:

```python
# 'ij,kl->ijkl'
for i in range(di):  
    for j in range(dj):   
        for k in range(dk):  
            for l in range(dl):  
                C[i,j,k,l] = A[i,j] * B[k,l] 

# 'ijk,jkl->il
for i in range(di):  
    for j in range(dj):   
        for k in range(dk):  
            for l in range(dl):  
                C[i,l] += A[i,j,k] * B[j,k,l]
```

Am I missing something about why ``flop_count`` adds an extra factor if an inner sum is used?

---

By the way, I've actually realised I was plotting the optimized *scaling* rather than flops, which while it shows very similar trends, allows optimal to have worse scaling than greedy but still better flops overall - so actually optimal is in fact always better as expected. Here is the corrected version:

opt_einsum_flops

dgasmith:
Right so `a * b` is one operation `c = c + a * b` takes two operations (without FMA). With FMA `c += a * b` is a single operation.

Thanks for the updated chart, it looks a bit more inline with expectations!

jcmgray:
Right, that makes sense now. Thanks for those detailed explanations!

jcmgray:
Woops so one more update to the FLOPS graph - I had only update the scatter points and not the lines. Here is the proper one (last update I hope!):

opt_einsum_flops

The good news, is that the flops speedups are actually much more pronounced.

dgasmith:
It looks like there are several cases where `greedy` does better than `branch-1`. How costly would it be to run both algorithms and take the best for up to 10 tensors or so, perhaps a factor of 2?

jcmgray:
OK updated timings:

opt_einsum_timings

And flops speedup:

opt_einsum_flops

I am a bit surprised that ``branch-1`` always seems to be better than ``greedy/eager`` - but this is maybe attributable to the fact that it sorts with a tuple of 'cost' (``size12 - size1 -size2``) as well as flops? -- Should we make ``auto`` use ``branch-1`` up until ``n~13``?

Also any preferences @dgasmith about what to do about regarding ``greedy + memory_limit``?

dgasmith:
Can you plot speedup as log? That might make the graphs a bit more appropriate.

`greedy` was turned into `eager` IIRC, can you check that the old greedy has the same performance as eager?

jcmgray:
> Can you plot speedup as log? That might make the graphs a bit more appropriate.

Sure thing (I've shown the errors on the mean as well):

opt_einsum_flops

> greedy was turned into eager IIRC, can you check that the old greedy has the same performance as eager?

Yes I found ``greedy`` had very similar characteristics to ``eager`` in terms of contraction quality (maybe even slightly worse for large contractions - but this could have been noise). They are doing essentially the same thing so this makes sense.

dgasmith:
@fritzo Any comments on this?

On the `greedy` comment. If the code is just slower with `memory_limit` then I think it is fine, perhaps add a note in the docs.

I do not see any further blockers and I think we should move this in. Be good to split out the speedup discussion to another issue so that we can look at it in detail.

jcmgray:
I've updated the docs quite a bit for these changes - including merging ``eager``/``greedy``. Any thoughts/comments regarding phrasing etc obviously welcome!

dgasmith:
Can we change that "Path Finding" page to introduction or similar and see if see if we can squeeze some of the above charts into that page which might give people a bit more of a visual illustration of why we have so many paths. We also might want to reorder the path documents as well so we go from most costly to least costly.

This is a lot of work, thanks for really pushing this forward!

jcmgray:
Will make those changes soon just a bit busy at the moment!

jcmgray:
OK I have made a few updates to this.

1. Changed the docs to 'introduction' and added a 'performance comparison' section with the graphs.
2. Re-ordered the sections into decreasing cost.
3. Added ``'branch-1'`` as an option and made ``'auto'`` mode use it for ``<= 14`` inputs. Since it explicitly considers the flop cost as each step it can be marginally higher quality for small-medium contractions than ``'greedy'`` (see graphs). I expect it will not be used much outside of auto mode.
4. Added a few options to ``rand_equation``.

N.B. one final note about the speedup graphs. I realised while doing this that you have to take statistics on *log* speedup otherwise large speedups massively outweigh large slowdowns and you get a weird effect where two different algorithms seem to be on average faster than each other. I've corrected that now.

opt_einsum/paths.py
```diff
@@ -484,5 +497,11 @@ def eager(inputs, output, idx_dict):
...
     [(0, 2), (0, 1)]
     """
+    if memory_limit not in _UNLIMITED_MEM:
+        return branch(inputs, output, idx_dict, memory_limit, nbranch=1)
+
     ssa_path = _ssa_optimize(inputs, output, idx_dict)
     return ssa_to_linear(ssa_path)
+
+
+greedy = eager
```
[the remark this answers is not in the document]
jcmgray: I was just trying to maintain maximum backward compatibility. But yes probably cleaner to just rename ``eager->greedy``, unless @fritzo has any objections? ``'eager'`` will still be used as an alias with ``optimize='eager'`` but otherwise will not be referenced.

jcmgray:
[By the way, here's that build of the docs for these changes again.](https://oetest.readthedocs.io/en/recursive-optimal/path_finding.html)

> I need to give this another once over with fresh eyes. In the meantime, are there remaining action items that you would like to hit?

I don't think so! Depending on how you feel about the docs (which admittedly have quite a lot of information spread out now), this is good to go from my perspective. 

Would be nice to get into a release I think as there should be some nice automatic speedups for small to medium sized expressions.

fritzo:
This generally looks reasonable to me. Right now we're using a fork of the `eager` path in Pyro, but I think after this PR we'll switch to `auto` :smile: (as I understand, `auto` now uses the algorithm formerly known as `eager` on large contractions).

jcmgray:
> This generally looks reasonable to me. Right now we're using a fork of the eager path in Pyro, but I think after this PR we'll switch to auto 😄 (as I understand, auto will henceforth use the algorithm formerly known as eager on large contractions).

Correct!

dgasmith:
Pulling this in, overall this is a great change which gives us quite a few options in the future as well. Thanks again @jcmgray for another superb pull request to this project!

PR #78, opened 2018-12-11
random-greedy, PathOptimizer & cost_fn

### LINKED ISSUE #63: Use best of many random greedy paths

### LINKED ISSUE #64: Allow custom cost objective functions

jcmgray:
This is a work in progress for implementing heuristic random optimizers (closes #63) for contractions that are both large and tricky. Mainly it does this by implementing a ``choose_fn`` option that is in charge of choosing which contraction to take during ``greedy``, but it also adds some other infrastructure changes. These include:

- A ``PathOptimizer`` class which allows
    - user-defined optimizers
    - tweaking advanced settings of an optimizer (e.g. ``BranchOptimizer`` which is now a subclass)
    - persisting an optimizer (e.g. calling random-greedy until you are satisfied)
- A ``cost_fn`` option in branch and greedy for customising the heuristic used to assess contractions (closes #64)
- Allowing ``branch`` and ``random-greedy`` to minimize 'flops' or 'size'

Some details on the ``RandomOptimizer``:

- It can take a executor-pool instance so that the path-finding can be trivially parallelized/distributed. 
- You can specify the maximum amount of time to run for (making it an 'anytime' algorithm)
- It chooses contractions by weighting them according to a 'thermal' boltzmann distribution based on their associated heuristic cost. This makes sure that it branches intelligently - only when the possible options are quite similarly costed. 
- You can tweak how 'adventurous' to make it by adjusting the temperature.

For the default settings (``optimize='random-greedy'`` which does 32 repeats) here's the basic results for speedup:

download

So pretty handy.

Here's some sample usage of manually tweaking/running an optimizer:

```python
import opt_einsum as oe                                
import numpy  as np 
from concurrent.futures import ProcessPoolExecutor     

# set up a random problem
eq, shapes = oe.helpers.rand_equation(30, 3)           
views = [np.random.rand(*s) for s in shapes] 

# set up a parallel random optimizer instance
pool = ProcessPoolExecutor()                           
optimizer = oe.paths.RandomOptimizer(minimize='size', executor=pool)

# optimize the path in several steps
for temp in [100, 10, 1, 0.1]:  
    optimizer.temperature = temp  # can change temp between runs
    path_info = oe.contract_path(eq, *views, optimize=optimizer)[1] 

# now we can use the best path found over all runs
oe.contract(eq, *views, optimize=optimizer.path)                    
1.2705446680196601e+22
```

Some other notes
----------------------------

- ~~I've tried to modify ``greedy`` as little as possible but the additional options do make it a bit slower (5%-10%) for large (300+ tensors) contractions (cc @fritzo).~~ Neglible overhead now.
- For simple contractions (i.e. where the graph is ~1-dimensional) random-greedy does not offer much improvement.
- For v. tricky contractions, although the gains over greedy are still big, the final result is quite probably still far from optimal 
- Real world I've been testing this approach to replicate the results (with success) of e.g. this paper https://arxiv.org/abs/1811.09599
- For the ``cost_fn``, 'memory removed' does seem to be the best heuristic I've found still. However applying some slight randomization helps break ties and explore the full contraction space.

TODO
----

- [x] Decide on various names (@dgasmith)
- [x] Add to docs (including what ``cost_fn`` and ``choose_fn`` are somewhere)
- [x] Test custom BranchOptimizer and RandomOptimizer settings / reuse
- [x] Generate some results rationalizing the choice of default random-greedy settings
- [x] Modify ``cost_fn`` signature?
- [x] Add some preset ``cost_fn`` options?

I've tried to make all the different additions modular to keep them simple and extendable, but obviously there are some big changes here - feedback v. welcome!


opt_einsum/paths.py
```diff
@@ -193,154 +210,213 @@ def _optimal_iterate(path, remaining, inputs, flops):
...
-                i, j = j, i
-            k1, k2 = inputs[i], inputs[j]
+                # set cost heuristic in order to locally sort possible contractions
+                size1, size2 = size_cache[inputs[i]], size_cache[inputs[j]]
 
-            # initially ignore outer products
-            if k1.isdisjoint(k2):
-                continue
+                if self.cost_fn is None:
+                    cost = size12 - size1 - size2
```
dgasmith: Maybe add this cost function to the dictionary set if none in the init? Would simplify the logic here.

opt_einsum/paths.py
```diff
@@ -506,3 +596,257 @@ def greedy(inputs, output, size_dict, memory_limit=None):
...
+            # check if we have found a new best
+            found_new_best = self.better(cost, size, self.best['cost'], self.best['size'])
+
+            if found_new_best:
+                self.best['cost'] = cost
+                self.best['size'] = size
+                self.best['ssa_path'] = ssa_path
+
+            # check if we have run out of time
+            if (self.max_time is not None) and (time.time() > t0 + self.max_time):
```
dgasmith: `if self.executor is not None:` can be the exterior if here.

dgasmith:
A few comments on your notes:
 - Looking at the chart it seems like we really need to consider running `greedy` for all branch codes to catch the regression cases. I think greedy should be fast enough that at least for branch-2 it should be trivial to compute.
 - We could likely have some fun heuristics trying to estimate how much we gain by exploring more random paths vs how much time the contraction will take or how many random paths we can explore before we take ~1% of the contraction execution time.

Thanks for this pretty massive change! I look forward to diving into this a bit.

opt_einsum/paths.py
jcmgray: Yep, I've simplified the logic for the ``cost_fn`` and ``choose_fn`` [now](https://github.com/dgasmith/opt_einsum/pull/78/commits/18df4be675746daaa203c84e59f26bfdc15c92d8), which has also reduced any overhead to negligible.

opt_einsum/paths.py
jcmgray: Ah yes good point.

dgasmith:
I forget, did you post the code to generate the graphs you have been making? Perhaps the full code is not required, but a template for random paths and examination of speedups would be great. I am going to carve out some time to play with this a bit and see if I can get a better handle on the bounds.

jcmgray:
> I am still traveling (!) so I do not have too much time. Are there specific items I can help on here so that I am not holding you up?

Nothing in particular I don't think! I am also busy which is why I have been making rather slow progress. If the general approach here looks good to you then I will commit to writing some docs and the last few tests ready for final review.

dgasmith:
Yea, I am pretty happy with the general approach. It all look good to me!

jcmgray:
OK I have added a first run at the docs. With that I think everything is pretty ready now.

Just a note that the 1.54% coverage decrease is I think coming from ``pytorch.tensordot`` being picked up now rather than the custom implementation.

jcmgray:
Also, I should add that I've implemented a generic ``RandomOptimizer`` instance that you can subclass to make custom random path optimizers, without having to re-write all the stuff ``RandomGreedy`` is using such as the parallelization etc.

Part of the motivation for this is I have been playing with path optimizers that take a top-down approach of recursively partitioning the graph (as in [this paper](https://arxiv.org/abs/1805.00475)) - it works very well for some cases! Anyway, I think these will be too much to include in ``opt_einsum`` directly, but it would be nice to plug them in v easily.

docs/source/random_greedy_path.rst
```diff
@@ -0,0 +1,123 @@
...
+naturally suited to parallelization. This is achieved by supplying a
+pool-executor like object, which should have an api matching the python 3
+`concurrent.futures <https://docs.python.org/3/library/concurrent.futures.html>`_
+module:
+
+.. code:: python
+
+    from concurrent.futures import ProcessPoolExecutor
+
+    executor = ProcessPoolExecutor()
```
dgasmith: Here is there any reason not to use ThreadPoolExecutor? These should be self contained functions and the overhead is cheaper for thread pools.

opt_einsum/path_random.py
```diff
@@ -0,0 +1,320 @@
...
+        module ``concurrent.futures``. Namely, a ``submit`` method that returns
+        ``Future`` objects, themselves with ``result`` and ``cancel`` methods.
+    pre_dispatch : int, optional
+        If using an ``executor``, how many jobs to pre-dispatch so as to avoid
+        submitting all jobs at once. Should also be more than twice the number
+        of workers to avoid under-subscription. Default: 128.
+
+    Attributes
+    ----------
+    path : list[tuple[int]]
```
dgasmith: Reminds me that we should look into a Py2 deprecation plan and think about Py3+ (and typing!).

opt_einsum/path_random.py
```diff
@@ -0,0 +1,320 @@
...
+    Parameters
+    ----------
+    max_repeats : int, optional
+        The maximum number of repeat trials to have.
+    max_time : float, optional
+        The maximum amount of time to run the algorithm for.
+    minimize : {'flops', 'size'}, optional
+        Whether to favour paths that minimize the total estimated flop-count or
+        the size of the largest intermediate created.
+    executor : executor-pool like, optional
```
dgasmith: Could `executor` be more like:
- `threading=True` which would build a ThreadPool
- `int` which would pick the number of threads to use
- `ThreadPool` to use

Might make it a bit easier on the user.

opt_einsum/path_random.py
```diff
@@ -0,0 +1,320 @@
...
+
+    def _gen_results_executor(self, repeats, trial_fn, args):
+        """Lazily generate results from an executor without submitting all jobs at once.
+        """
+        self._futures = deque()
+
+        for r in repeats:
+            if len(self._futures) < self.pre_dispatch:
+                self._futures.append(self.executor.submit(trial_fn, r, *args))
+                continue
```
dgasmith: Do you need this continue?

opt_einsum/paths.py
```diff
@@ -193,154 +220,230 @@ def _optimal_iterate(path, remaining, inputs, flops):
...
+                    if candidate:
+                        heapq.heappush(candidates, candidate)
+
+            # recurse into all or some of the best candidate contractions
+            bi = 0
+            while (self.nbranch is None or bi < self.nbranch) and candidates:
+                _, _, new_flops, new_size, (i, j), k12 = heapq.heappop(candidates)
+                _branch_iterate(path=path + ((i, j),),
+                                inputs=inputs + (k12,),
+                                remaining=remaining - {i, j} | {len(inputs)},
```
dgasmith: Do we need to worry slightly about order of operations here? I try to be fairly specific just so it is very clear to the user.

opt_einsum/contract.py
```diff
@@ -31,15 +32,21 @@ class PathInfo(object):
...
-                 indices, scale_list, naive_cost, opt_cost, size_list):
+    def __init__(self, contraction_list, input_subscripts, output_subscript, indices,
+                 path, scale_list, naive_cost, opt_cost, size_list, size_dict):
         self.contraction_list = contraction_list
         self.input_subscripts = input_subscripts
         self.output_subscript = output_subscript
+        self.path = path
         self.indices = indices
         self.scale_list = scale_list
         self.naive_cost = naive_cost
```
dgasmith: Maybe save `opt_cost` and `naive_cost` as Decimal and then add the speedup here as we seem to access it in the docs a bit? Should be fine to import decimal overall as it is quite lightweight.

docs/source/random_greedy_path.rst
```diff
@@ -0,0 +1,123 @@
...
+whilst reusing it. We'll also set a high ``max_repeats`` and instead use a
+maximum time to terminate the search:
+
+.. code:: python
+
+    optimizer = oe.RandomGreedy(max_time=2, max_repeats=1_000_000)
+
+    for T in [1000, 100, 10, 1, 0.1]:
+        optimizer.temperature = T
+        path_rand_greedy = oe.contract_path(eq, *arrays, optimize=optimizer)[1]
```
dgasmith: Could you potentially print out the best cost at each iteration? Could be interesting for demonstration.

opt_einsum/path_random.py
```diff
@@ -0,0 +1,320 @@
...
+
+    # compute relative probability for each potential contraction
+    if temperature == 0.0:
+        energies = [1 if cost == cmin else 0 for cost in costs]
+    else:
+        # shift by cmin for numerical reasons
+        energies = [math.exp(-(c - cmin) / temperature) for c in costs]
+
+    # randomly choose a contraction based on energies
+    chosen, = random.choices(range(n), weights=energies, k=1)
```
dgasmith: You can remove `k=1` as it defaults.

opt_einsum/path_random.py
```diff
@@ -0,0 +1,320 @@
...
+    # adjust by the overall scale to account for fluctuating absolute costs
+    if rel_temperature:
+        temperature *= max(1, abs(cmin))
+
+    # compute relative probability for each potential contraction
+    if temperature == 0.0:
+        energies = [1 if cost == cmin else 0 for cost in costs]
+    else:
+        # shift by cmin for numerical reasons
+        energies = [math.exp(-(c - cmin) / temperature) for c in costs]
```
dgasmith: Worth using NumPy here? Might save a bit of time/be slightly cleaner.

opt_einsum/path_random.py
```diff
@@ -0,0 +1,320 @@
...
+    def _gen_results_executor(self, repeats, trial_fn, args):
+        """Lazily generate results from an executor without submitting all jobs at once.
+        """
+        self._futures = deque()
+
+        for r in repeats:
+            if len(self._futures) < self.pre_dispatch:
+                self._futures.append(self.executor.submit(trial_fn, r, *args))
+                continue
+            yield self._futures.popleft().result()
```
dgasmith: Why two yields here? I think only should need the loop.

opt_einsum/path_random.py
```diff
@@ -0,0 +1,320 @@
+import math
+import time
+import heapq
+import random
+import functools
+from collections import deque
+
+from . import helpers
+from .paths import PathOptimizer, _ssa_optimize, _BETTER_FNS, ssa_to_linear, _calc_k12_flops
```
dgasmith: Usually we try not to import `_*` functions. Maybe convert add a getter function to `_BETTER_FNS` and move those others to non `_`?

opt_einsum/path_random.py
```diff
@@ -0,0 +1,320 @@
+import math
```
dgasmith: Can you add a small comment header at the top?

docs/source/random_greedy_path.rst
jcmgray: The problem is the GIL. Since greedy for example is essentially pure python, there is no meaningful parallelism unless we use processes. On the other hand, I have a prototype optimizer that uses [METIS](https://metis.readthedocs.io/en/latest/), and there, assuming most time is spent in the c-code, it would make sense to use a thread pool.

opt_einsum/path_random.py
jcmgray: Yes, I could definitely add ``executor=True`` and ``executor=int`` options, such that a ``concurrent.futures.ProcessPoolExecutor`` is created automatically with either the default or specified number of workers. I feel like ``executor`` is the most precise name, but we could also think about changing the keyword to ``parallel`` or something?

opt_einsum/path_random.py
jcmgray: So we need the ``continue`` and the second ``yield`` below in conjunction - the continue means that ``pre_dispatch`` number of jobs are submitted before we start yielding futures (otherwise only 1 would). The second yield is to then exhaust the queue once we have submitted all the jobs.

opt_einsum/paths.py
jcmgray: Not sure I quite understand - could you elaborate?

opt_einsum/contract.py
jcmgray: Yes, will do.

docs/source/random_greedy_path.rst
jcmgray: That would be nice - I'll add.

opt_einsum/path_random.py
jcmgray: I'll check! But I think given the small size of the list (defaults to 8) the constant overhead of initializing ``numpy`` arrays probably makes it slower.

docs/source/random_greedy_path.rst
dgasmith: Sorry, I always forget what the GIL does and does not lock as we usually thread in other languages. If the inputs are independent and/or copied, shouldn't the GIL release?

opt_einsum/path_random.py
dgasmith: What about `nthread`? I suspect that most folks would let the function build its own executor.

opt_einsum/paths.py
dgasmith: `remaining - ({i, j} | {len(inputs)})`

jcmgray: Ah I see, I'll add some parentheses.

docs/source/random_greedy_path.rst
jcmgray: Sadly no, GIL is by default always on even if all the data structures should be thread-safe. Its only really calls into other languages and IO etc. that threads offer speed-up for.

opt_einsum/path_random.py
dgasmith: Ah, gotcha. Makes sense, could you add a comment or two explaining this.

opt_einsum/path_random.py
jcmgray: Given above comments about threading, maybe ``parallel`` works best as a compromise? I think the three possible use cases all work reasonably well:
```python
RandomGreedy(parallel=True)  # probably the most used
RandomGreedy(parallel=4)
RandomGreedy(parallel=pool)
```
whereas something like ``nproc=pool`` might be a bit confusing.

dgasmith: Sounds good and I agree that `parallel=True` is likely the most common case here.

jcmgray:
> Whew, that was a PR.

Yes this did get rather big - I appreciate the careful review! I've now implemented ``RandomGreedy(parallel=True)``, and I think addressed all the other comments? So good to go from my end.

dgasmith:
We look good! Is this ready?

Issue #46, opened 2018-08-22
NumPy Dispatch Mechanisms


dgasmith:
It would be a good idea to keep an eye on [NEP 18](https://www.numpy.org/neps/nep-0018-array-function-protocol.html) "A dispatch mechanism for NumPy’s high-level array." Depending on the implementation of the NEP, this could help us avoid extra code or may require us to rework how our backend technology works.

dgasmith:
This is making progress in the NumPy ecosystem:
https://github.com/numpy/numpy/issues/12028
https://github.com/numpy/numpy/pull/12163

fritzo:
I'm excited for NEP 18 to go through, but at the same time, I really like that opt_einsum provides a pure-python dispatch mechanism that we can essentially use for symbolic computation. I've opened #90 to explain some of these use cases.

jcmgray:
For what its worth, I expanded the dispatch mechanism here into [autoray](https://github.com/jcmgray/autoray). I was kind of surprised nothing like that existed already (possibly I just couldn't find it!). Anyway, ``opt_einsum`` probably doesn't need another dependency like ``autoray`` just for ``tensordot`` and friends, but it might be of interest anyway.

dgasmith:
It would be good to do, but not high on my feature set.

PR #81, opened 2019-02-16
Implement backend='auto' and make it the default

jcmgray:
## Description
This implements ``backend='auto'`` for contractions and makes it the default. This works currently by inferring the module from the class of the first array supplied. A slight detail is that sometimes the module of an array does not supply ``tensordot`` etc. (maybe it has subclassed just ``numpy.ndarray`` for example) and in this case ``'numpy'`` is still used.

Hopefully this kind of thing will eventually be made redundant by ``__array_function__`` (#46) but for the moment it makes it a bit more convenient to work with other array libraries.

## Status
- [x] Ready to go

### DIFF
--- a/opt_einsum/contract.py
+++ b/opt_einsum/contract.py
@@ -477,7 +477,7 @@ def contract(*operands, **kwargs):
     # Grab non-einsum kwargs
     use_blas = kwargs.pop('use_blas', True)
     memory_limit = kwargs.pop('memory_limit', None)
-    backend = kwargs.pop('backend', 'numpy')
+    backend = kwargs.pop('backend', 'auto')
     gen_expression = kwargs.pop('_gen_expression', False)
     constants_dict = kwargs.pop('_constants_dict', {})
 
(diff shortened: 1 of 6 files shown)


opt_einsum/contract.py
```diff
@@ -500,14 +500,35 @@ def contract(*operands, **kwargs):
...
+    """Find out what backend we should use, dipatching based on the first
+    array if ``backend='auto'`` is specified.
+    """
+    if backend != 'auto':
+        return backend
+    backend = infer_backend(arrays[0])
+
+    # some arrays will be defined in modules that don't implement tensordot
+    # etc. so instead default to numpy
+    if not backends.has_tensordot(backend):
```
dgasmith: What if someone has subclassed a NumPy array or similar or uses an unknown backend? We should fall back to NumPy here as well.

jcmgray: So I _think_ the behavior you describe is what happens here. For example, in my library ``quimb`` I have a subclass of ``ndarray`` so the initial inferred backend is ``quimb``. However this check for ``quimb.tensordot`` (which is cached) means ``'numpy'`` is used still since unlike dask, tensorflow etc. ``quimb`` does not provide tensordot.

jcmgray:
> I like the idea of this, but worried that this could potentially break code. We seem to be in quite a few projects at the moment and could change behavior.

Yeah I definitely appreciate this might be a significant change and would be good to get anybody's feedback! I think for almost all use cases there should be no change in behaviour however. 

The only really 'breaking' edge case I can think of, is if someone is using non-numpy arrays from a library that *does* provide ``tensordot`` but also is also for some reason relying by default on ``numpy.tensordot`` being the backend function. In this case, the libraries ``tensordot`` will now be used instead.

fritzo:
IIUC this would not break any Pyro code since we always specify backend, and actually we usually specify a custom backend like `pyro.ops.einsum.torch_log`.

jcmgray:
> IIUC this would not break any Pyro code since we always specify backend, and actually we usually specify a custom backend like `pyro.ops.einsum.torch_log`.

Yes no change for explicit ``backend`` - which I imagine is all uses in other projects since this has until now been essentially required.

---

The overall logic is really that if you supply a ``mylib`` array, and ``mylib`` implements tensor functions, you almost definitely want to use that backend and so supplying ``backend='mylib'`` feels redundant. Of course, if you really _wanted_ to use ``numpy`` tensor functions you can explicitly set ``backend='numpy'`` but since this is the much more niche case it feels fine to be more verbose.

dgasmith:
I think this represents a fundamental shift where we are no longer focused on a NumPy backend, but are agnostic. I would consider this a major change (3.x) as we change our philosophy statement (something that could be argued that we have done anyways). This would make me a bit more comfortable as we are potentially breaking code, thoughts?

jcmgray:
Yes that seems like a reasonable approach to me!

PR #84, opened 2019-02-20
Factor out auto path optimizer + allow shapes in contract_path

### LINKED ISSUE #82: FR: separate 'auto' strategy out of contract_path
The 'auto' strategy is really nice, but it requires actual arrays because it is based into `contract_path`. Sometimes I want to get a path based merely on shapes. It should be easy to factor out the logic from `contract_path` to a new `paths.auto` that conforms to the standard shape-based path interface.

jcmgray:
## Description
A couple of minor-ish path tweaks:
 1. Factor out the ``'auto'`` path logic to an actual path function (closes #82) cc @fritzo 
 2. Allow ``contract_path`` to accept shapes with the option ``shapes=True`` (rather than requiring actual array-like objects as now)

The path options and auto path choosing based on size I've implemented as dictionaries, which as well as for efficiencies sake potentially allows them to be customized in the future. E.g. users adding their own named optimizers options or setting custom 'auto' strategies.

## Status
- [x] Ready to go

### DIFF
--- a/opt_einsum/path_random.py
+++ b/opt_einsum/path_random.py
@@ -372,6 +372,3 @@ def random_greedy(inputs, output, idx_dict, memory_limit=None, **optimizer_kwarg
     """
     optimizer = RandomGreedy(**optimizer_kwargs)
     return optimizer(inputs, output, idx_dict, memory_limit)
-
-
-paths.add_path_fn('random-greedy', random_greedy)
--- a/opt_einsum/path_random.py
+++ b/opt_einsum/path_random.py
@@ -11,13 +11,13 @@ import functools
 from collections import deque
 
 from . import helpers
-from .paths import PathOptimizer, ssa_greedy_optimize, get_better_fn, ssa_to_linear, calc_k12_flops
+from . import paths
 
 
 __all__ = ["RandomGreedy", "random_greedy"]
 
 
-class RandomOptimizer(PathOptimizer):
+class RandomOptimizer(paths.PathOptimizer):
     """Base class for running any random path finder that benefits
     from repeated calling, possibly in a parallel fashion. Custom random
     optimizers should subclass this, and the ``setup`` method should be
@@ -81,7 +81,7 @@ class RandomOptimizer(PathOptimizer):
         self.max_repeats = max_repeats
         self.max_time = max_time
         self.minimize = minimize
-        self.better = get_better_fn(minimize)
+        self.better = paths.get_better_fn(minimize)
         self.parallel = parallel
         self.pre_dispatch = pre_dispatch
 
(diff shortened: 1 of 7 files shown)


opt_einsum/paths.py
```diff
@@ -645,3 +652,50 @@ def greedy(inputs, output, size_dict, memory_limit=None, choose_fn=None, cost_fn
...
+    'greedy': greedy,
+    'eager': greedy,
+    'opportunistic': greedy,
+}
+
+
+def add_path_fn(name, fn):
+    """Add path finding function ``fn`` as an option with ``name``.
+    """
+    _PATH_OPTIONS[name] = fn
```
dgasmith: Can we 1) lowercase name and 2) make sure the key does not already exist?

opt_einsum/path_random.py
```diff
@@ -372,3 +372,6 @@ def random_greedy(inputs, output, idx_dict, memory_limit=None, **optimizer_kwarg
     """
     optimizer = RandomGreedy(**optimizer_kwargs)
     return optimizer(inputs, output, idx_dict, memory_limit)
+
+
+paths.add_path_fn('random-greedy', random_greedy)
```
dgasmith: I do like the registration, but the potential circular imports worry me a bit. What do you think of imports the path finders into `paths.py` and register them there?

opt_einsum/paths.py
```diff
@@ -645,3 +652,50 @@ def greedy(inputs, output, size_dict, memory_limit=None, choose_fn=None, cost_fn
...
+    'branch-all': branch_all,
+    'branch-2': branch_2,
+    'branch-1': branch_1,
+    'greedy': greedy,
+    'eager': greedy,
+    'opportunistic': greedy,
+}
+
+
+def add_path_fn(name, fn):
```
dgasmith: Could we call this `register_path_fn`, would make the language slightly clearer I believe.

opt_einsum/path_random.py
jcmgray: So the registration was my way of getting round the circular import problem since ``path_random`` depends on ``paths`` quite a bit. However switching to ``from . import paths`` in this file might have sorted that already. Alternatively, could perform the registration in the top ``__init__``?

dgasmith: The `from . imports paths` did prevent the circular depends in this particular case, but is quite vulnerable to problems in the future. Perhaps we could split off registration into another file or as you said, add it to the init.

jcmgray:
I've made those changes @dgasmith and moved the random-greedy registration into the top ``__init__``.

opt_einsum/paths.py
```diff
@@ -645,3 +652,53 @@ def greedy(inputs, output, size_dict, memory_limit=None, choose_fn=None, cost_fn
...
+    'branch-all': branch_all,
+    'branch-2': branch_2,
+    'branch-1': branch_1,
+    'greedy': greedy,
+    'eager': greedy,
+    'opportunistic': greedy,
+}
+
+
+def register_path_fn(name, fn, overwrite=False):
```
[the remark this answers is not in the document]
jcmgray: Sure thing, I was thinking here one might want to update a function, but this can still be explicitly done by deleting the key first.

Issue #99, opened 2019-08-22
`Optimal` path not optimal?


Bonnevie:
First, thanks for the great package!

Was running `optimal` vs. `auto` on a 10 element random expression, and while I know optimal is not recommended for larger networks, I found it odd that it gives worse scaling and with a larger intermediate size. Some of the other threads mention memory as a factor in the optimization, but the `auto` solution seems better on memory as well? All the tensors are 3x3. Is this intentional behaviour? 

auto path:
```
([(2, 3), (4, 8), (0, 4), (0, 4), (4, 5), (0, 2), (2, 3), (1, 2), (0, 1)],
   Complete contraction:  db,cc,fe,fe,aa,ff,fe,cb,ea,ac->d
          Naive scaling:  6
      Optimized scaling:  3
       Naive FLOP count:  7.290e+3
   Optimized FLOP count:  2.700e+2
    Theoretical speedup:  27.000
   Largest intermediate:  9.000e+0 elements
 --------------------------------------------------------------------------------
 scaling        BLAS                current                             remaining
 --------------------------------------------------------------------------------
    2              0              fe,fe->fe         db,cc,aa,ff,fe,cb,ea,ac,fe->d
    2              0              fe,fe->fe            db,cc,aa,ff,cb,ea,ac,fe->d
    3           GEMM              cb,db->cd               cc,aa,ff,ea,ac,fe,cd->d
    2              0              ac,cc->ac                  aa,ff,ea,fe,cd,ac->d
    3           GEMM              ac,cd->ad                     aa,ff,ea,fe,ad->d
    2              0              ea,aa->ea                        ff,fe,ad,ea->d
    3           GEMM              ea,ad->ed                           ff,fe,ed->d
    3           GEMM              ed,fe->df                              ff,df->d
```

optimal path: 
```
([(1, 7), (0, 8), (0, 2), (0, 1), (0, 5), (0, 3), (0, 3), (0, 2), (0, 1)],
   Complete contraction:  db,cc,fe,fe,aa,ff,fe,cb,ea,ac->d
          Naive scaling:  6
      Optimized scaling:  4
       Naive FLOP count:  7.290e+3
   Optimized FLOP count:  5.130e+2
    Theoretical speedup:  14.211
   Largest intermediate:  2.700e+1 elements
 --------------------------------------------------------------------------------
 scaling        BLAS                current                             remaining
 --------------------------------------------------------------------------------
[...]

PR #102, opened 2019-08-28
added dynamic programming path

mrader1248:
## Description
In the field of tensor network states (matrix product states, projected entangled pair states, ...) the most common approach to optimise tensor network contractions is based on dynamic programming [Phys. Rev. E 90, 033315 (2014)], which usually provides an optimal contraction path within a reasonable amount of time (even for a large number of tensors). This pull request provides such an implementation.

## Questions
- [ ] Is there a simple way to run the benchmarks of the "Performance Comparison" section of the documentation for the dynamic programming optimiser to compare with the other approaches?


opt_einsum/paths.py
```diff
@@ -652,6 +655,131 @@ def greedy(inputs, output, size_dict, memory_limit=None, choose_fn=None, cost_fn
...
+            c.append(i)
+    
+    return s
+
+
+class DynamicProgrammingOptimizer(PathOptimizer):
+    
+    def __call__(self, inputs, output, size_dict, memory_limit=None):
+        
+        union = lambda a: functools.reduce(lambda x, y: x | y, a, set())
```
jcmgray: Python sets do have a ``set.union(*my_sets)`` method, though won't work for length 0 ``my_sets``.

opt_einsum/paths.py
```diff
@@ -652,6 +655,131 @@ def greedy(inputs, output, size_dict, memory_limit=None, choose_fn=None, cost_fn
...
+        prod = lambda a: functools.reduce(lambda x, y: x * y, a, 1)
```
jcmgray: You could use ``helpers.compute_size_by_dict(inds, size_dict)`` directly instead of this.

opt_einsum/paths.py
```diff
@@ -652,6 +655,131 @@ def greedy(inputs, output, size_dict, memory_limit=None, choose_fn=None, cost_fn
...
+        
+        for n in range(2, len(inputs) + 1):
+            # compute x[n]
+            x.append(dict())
+            
+            # try to combine solutions from x[m] and x[n-m]
+            for m in range(1, n // 2 + 1):
+                for s1, (i1, cost1, cntrct1) in x[m].items():
+                    for s2, (i2, cost2, cntrct2) in x[n-m].items():
+                        if len(s1 & s2) == 0: # s1 and s2 have to be disjoint
```
jcmgray: Python sets also have a handy (and quicker) ``s1.isdisjoint(s2)`` method.

opt_einsum/paths.py
```diff
@@ -652,6 +655,131 @@ def greedy(inputs, output, size_dict, memory_limit=None, choose_fn=None, cost_fn
...
+                            # avoid e.g. s1={0}, s2={1} and s1={1}, s2={0}
+                            if len(s1) != len(s2) or min(s1) < min(s2): 
+                            
+                                # ignore purely outer products: either tensor sets s1 and s2 are connected by a 
+                                # non-output index or none of them has remaining summation indices left
+                                if len((i1 & i2) - output) > 0 or (i1 & output == i1 and i2 & output == i2):
+                                    cost = cost1 + cost2 + prod(size_dict[j] for j in i1 | i2)
+                                    s = s1 | s2
+                                    if s not in x[n] or cost < x[n][s][1]:
+                                        r = set(range(len(inputs))) - (s1 | s2) # remaining set of tensors
```
jcmgray: Real micro-optimization stuff, but there's a few computations you don't need to repeat here:
 * ``s1 | s2``
 * ``(i1 & i2) - output``
 * ``i1 | i2``

opt_einsum/paths.py
mrader1248: Great, thanks.

opt_einsum/paths.py
mrader1248: Yes, I know, but it was not really clear to me, which version is better for readability. Nevertheless, I removed my union.

opt_einsum/paths.py
mrader1248: Okay, good to know. I replaced my calls to `prod`.

opt_einsum/paths.py
mrader1248: You are right. I didn't really optimise this code up to now.

mrader1248:
This code should work for contractions with summation indices occurring three or more times (although I am not entirely sure, how optimal the resulting contractions will be in general) and it also works for non-summation indices occuring more than once.

Do you have standard test cases you are using for the other optimisers as well to ensure they are yielding correct results?

It would be great, if you could run the performance comparison.

jcmgray:
> Do you have standard test cases you are using for the other optimisers as well to ensure they are yielding correct results?

Yes, check the ``test_contract.py`` and ``test_paths.py`` files - should just be a matter of adding a few parametrizations to the existing tests.

Those tweaks seem to have sped it up a fair bit - if you have any other potential optimizations in mind let me know, else I will run the benchmarks in a bit!

mrader1248:
I added `'dp'` to `all_optimizers` to test my implementation. You can still wait with running the benchmark until there are no more failures.
Something I just realised: The code in `contract` does not 'repair' the output indices before calling an optimiser? At the moment test cases like `"ab,ab,c->"` fail for my implementation as it assumes complete specification of the output indices (it works for `"ab,ab,c->c"` and `"ab,ab,c->abc"`). Do you have some common routine to repair the output indices?

jcmgray:
Not quite sure what you mean by 'repair', ``"ab,ab,c"`` would generate the output automatically as ``"ab,ab,c->c"``, but when the arrow is explicitly there ``"ab,ab,c->"`` it implies that the ``'c'`` dimension is indeed summed over and won't appear in the output, so this is still a valid contraction.

mrader1248:
Thank you for explaining this example to me. Indeed, I thought `"ab,ab,c"` and `"ab,ab,c->"` is the same. With 'repair' I meant that `contract` should transform `"ab,ab,c"` into `"ab,ab,c->c"` before passing it to the optimiser (and I just realised that this is actually done). So, nevermind.

Indeed, my code fails for summation indices occurring only once. However, as far as I understand it, summing over these indices in the beginning is always optimal?

jcmgray:
Here's the initial results I have for path finding time:

oe-dp-times

And FLOPS improvement over ``greedy``:

oe-dp-flops

For this class of problems (basically random regular like contractions) it seems to give similar quality to ``branch-all`` but is faster for 10+ tensors.

mrader1248:
> When you say "summation indices occurring only once" do you refer to `"ab,ab,c"` like operations in all cases? If so, I believe this is true and something that we have been toying around with as well. See [here](https://github.com/dgasmith/opt_einsum/issues/99).
> 
> Not exactly identical, but similar problems. I do believe completing all Hadamard products and forming all scalar values first will always be an optimal strategy.

It seems very reasonable that first summing over indices occurring only in one input (e.g. `"c"` and `"d"` for `"ab,ab,c,dd->"`) is indeed optimal (at least I cannot think of any counter example). I put the code achieving this in a function `_sum_single_tensor_indices(inputs, output)` if anybody wants to use it.

@jcmgray: Thank you for these first benchmarks. I realised there might be one more optimisation I should implement (decomposing the contraction graph into disconntected subgraphs first).

mrader1248:
I added another improvement, which causes the optimiser to scale only linearly with the number of disconnected subgraphs of the contraction graph. If all the tests are successful for the latest commit, it would be interesting to see if/how performance improved.

So far I only added the `'dp'` path to the tests in `test_contract.py`. What kind of tests have to be added in `test_paths.py`?

opt_einsum/paths.py
```diff
@@ -652,6 +655,202 @@ def greedy(inputs, output, size_dict, memory_limit=None, choose_fn=None, cost_fn
...
+        
+        for i in [i for i in j if type(i) != int]:
+            s[0] += (len(t) + len(c),)
+            c.append(i)
+    
+    return s
+
+
+def _sum_single_tensor_indices(inputs, output):
+
```
jcmgray: Could you document what the output etc of this is? As might be useful elsewhere!

mrader1248:
> With regard to `test_paths.py`, maybe just add it to the `path_edge_tests` cases to make sure it matches `optimal` and maybe to `test_can_optimize_outer_products` as well.

The problem with `path_edge_tests` is that the solutions are not unique. Considering the first test `'eb,cb,fb->cef'` (apparently it's tested with shapes `[(4, 3), (4, 3), (3, 3)]`) it doesn't matter if the contraction sequence is `((0, 2), (0, 1))` or `((1, 2), (0, 1))` (all your optimisers yield the first solution, my optimiser yields the second one).

The test `test_can_optimize_outer_products` has the same issue: The result is not unique. For `"ab,cd,ef,fg"` it doesn't matter if the summation over `f` is performed before or after the first outer product. However, at this point I realised something different as well: For an outer product of three tensors, I thought the sequence `[(0,1,2)]` should be used instead of `[(0,1), (0,1)]`. However, the latter version is twice as fast. Do you know why?

> Might it also be possible to lint it for PEP8 as well? (@dgasmith not sure if this is official standard for `opt_einsum` but certainly the rest of the codebase seems to conform currently).

Can you tell me where I violate PEP8?

opt_einsum/paths.py
mrader1248: Sorry that I have forgotten to document this function. However, I just found a bug in my code, so maybe this function will change anyway. (Nevertheless, the final version will have a short documentation.)

jcmgray:
> The problem with path_edge_tests is that the solutions are not unique. Considering the first test 'eb,cb,fb->cef' (apparently it's tested with shapes [(4, 3), (4, 3), (3, 3)]) it doesn't matter if the contraction sequence is ((0, 2), (0, 1)) or ((1, 2), (0, 1)) (all your optimisers yield the first solution, my optimiser yields the second one).

Yes I think these can just be thought of as regression tests, I'd just put in the path that ``'dp'`` currently finds, but personally also happy if you want to skip. By the way, is the path finding deterministic?

> For an outer product of three tensors, I thought the sequence [(0,1,2)] should be used instead of [(0,1), (0,1)]. However, the latter version is twice as fast. Do you know why?

I think in general, a pairwise contraction sequence always is better. For the first path you have ``2 * d^3`` operations, and in the second case ``d^2 + d^3`` operations. Moreover, in practical cases, the pairwise paths can use tensordot and BLAS, whereas if there are more than 2 tensors, ``einsum`` must be used.

> Can you tell me where I violate PEP8?

There are quite a few, but potentially just ones that are turned off in your editor! Here's what I get:
```
paths.py:16:73: W291 trailing whitespace
paths.py:662:6: W291 trailing whitespace
paths.py:669:6: W291 trailing whitespace
paths.py:670:79: W291 trailing whitespace
paths.py:673:1: W293 blank line contains whitespace
paths.py:676:1: W293 blank line contains whitespace
paths.py:677:12: E261 at least two spaces before inline comment
paths.py:680:1: W293 blank line contains whitespace
paths.py:684:1: W293 blank line contains whitespace
paths.py:688:1: W293 blank line contains whitespace
paths.py:692:1: W293 blank line contains whitespace
paths.py:697:1: W293 blank line contains whitespace
paths.py:700:1: W293 blank line contains whitespace
paths.py:703:1: W293 blank line contains whitespace
paths.py:706:20: E272 multiple spaces before keyword
paths.py:706:50: E271 multiple spaces after keyword
paths.py:708:1: W293 blank line contains whitespace
paths.py:717:1: W293 blank line contains whitespace
paths.py:725:1: W293 blank line contains whitespace
paths.py:726:60: E226 missing whitespace around arithmetic operator
paths.py:728:1: W293 blank line contains whitespace
paths.py:729:40: E261 at least two spaces before inline comment
paths.py:730:1: W293 blank line contains whitespace
paths.py:741:1: W293 blank line contains whitespace
paths.py:743:1: W293 blank line contains whitespace
paths.py:748:1: W293 blank line contains whitespace
paths.py:752:1: W293 blank line contains whitespace
paths.py:755:1: W293 blank line contains whitespace
paths.py:759:22: E261 at least two spaces before inline comment
paths.py:762:1: W293 blank line contains whitespace
paths.py:765:1: W293 blank line contains whitespace
paths.py:771:1: W293 blank line contains whitespace
paths.py:773:76: W291 trailing whitespace
paths.py:774:1: W293 blank line contains whitespace
paths.py:775:116: W291 trailing whitespace
paths.py:783:54: E261 at least two spaces before inline comment
paths.py:785:81: E261 at least two spaces before inline comment
paths.py:790:1: W293 blank line contains whitespace
paths.py:792:1: W293 blank line contains whitespace
paths.py:802:79: W291 trailing whitespace
paths.py:803:63: W291 trailing whitespace
paths.py:804:77: W291 trailing whitespace
paths.py:806:71: W291 trailing whitespace
paths.py:808:1: W293 blank line contains whitespace
paths.py:810:71: W291 trailing whitespace
paths.py:811:72: W291 trailing whitespace
paths.py:849:1: W293 blank line contains whitespace
```

mrader1248:
> Yes I think these can just be thought of as regression tests, I'd just put in the path that `'dp'` currently finds, but personally also happy if you want to skip. By the way, is the path finding deterministic?

Yes, the path finding is deterministic. I personally just don't like these kind of tests: If somebody changes the path finding function in the future in a way that will affect the resulting path, it will take an annoying amount of time to find out that actually the test requires a specific solution.

> I think in general, a pairwise contraction sequence always is better. For the first path you have `2 * d^3` operations, and in the second case `d^2 + d^3` operations. Moreover, in practical cases, the pairwise paths can use tensordot and BLAS, whereas if there are more than 2 tensors, `einsum` must be used.

Why is it `2 * d^3`? I thought, it should be `d^3` (looping over three indices and multiplying three numbers each iteration)? However, I see that for pairwise outer products you can use BLAS and BLAS of course always wins. ;-)

Ok, I will take a look at these PEP8 violations.

jcmgray:
> Yes, the path finding is deterministic. I personally just don't like these kind of tests: If somebody changes the path finding function in the future in a way that will affect the resulting path, it will take an annoying amount of time to find out that actually the test requires a specific solution.

Yes I can appreciate this reasoning - I guess ideally we would test against equivalent paths, but for the moment best to leave.

> Why is it 2 * d^3? I thought, it should be d^3 (looping over three indices and multiplying three numbers each iteration)? However, I see that for pairwise outer products you can use BLAS and BLAS of course always wins. ;-)

There is some subtlety here to do with compiler instructions etc but essentially ``opt_einsum`` assumes that multiplying 3 numbers is twice as expensive as multiplying two numbers (i.e. ``x * y * z == (x * y) * z``). Equivalently, if done that way ``(x * y)`` is recomputed for each ``d`` times redundantly.

mrader1248:
I fixed the bug I mentioned in the context of `_sum_single_tensor_indices`. This also led to the removal of this function.

I added the dp path to the tests in `test_paths.py` and used the results, my optimiser currently finds (and checked by hand that they are correct).

The performance of my optimiser might have slightly improved as I removed two set intersections from a condition as they are no longer needed due to the initial decomposition into disonnected subgraphs.

`pycodestyle --max-line-length=120 paths.py` now gives me no more errors.

One more thing I found out: Running the rather trivial contraction code
```
expr = oe.contract_expression("i,i,j,j,k,l->", *[v.shape]*6, optimize=...)
%timeit expr(*[v]*6)
```
gives me a runtime of ~2.2 ms for the optimal and ~1.3 ms for the dp path (for v with 1M float64 elements) and ~9 s (!) for the greedy path (for only 100k elements). The optimal path is `[(4,), (4,), (4, 5), (0, 1), (2, 3), (0, 1), (0, 1)]` and my dp path is `[(4,), (4,), (4, 5), (0, 1), (2, 3), (0, 1), (0, 1)]`. I don't really understand how my code can be faster than the optimal path (especially because the dp path has 7 instead of 5 elements and both should have require `4d` operations), but I thought this might be interesting for you.
Although this is quite off-topic now: The greedy path produces a `d^2`-solution which could be easily circumvented by performing all the summations over single inputs in the very beginning.

jcmgray:
Amazing, thanks for those changes. I will try and re-run the benchmarks soon, probably with all the other optimizers actually again since I have changed machines.

> I don't really understand how my code can be faster than the optimal path (especially because the dp path has 7 instead of 5 elements and both should have require 4d operations), but I thought this might be interesting for you.

The estimated flop cost, while well correlated with actual run time, is still just an estimate. There are many factors to do with copying, transposing, threadability ... etc that means some paths are just noticeably faster despite the same estimated cost.

> Although this is quite off-topic now: The greedy path produces a d^2-solution which could be easily circumvented by performing all the summations over single inputs in the very beginning.

I'm not sure I follow this, which bit of greedy? As far as I am aware the single d^2 step of greedy is to find all the initial inner products.

mrader1248:
> Amazing, thanks for those changes. I will try and re-run the benchmarks soon, probably with all the other optimizers actually again since I have changed machines.

Great. I am looking forward to the benchmark results.

> I'm not sure I follow this, which bit of greedy? As far as I am aware the single d^2 step of greedy is to find all the initial inner products.

With `d` I meant the length of `v`. The greedy path is `[(0, 1), (0, 1), (0, 1), (0, 1), (0, 1)]`. The first two operations are the two scalar products [`O(d)`] and the third operation is a sum over an outer product [`O(d^2)`].

jcmgray:
> With d I meant the length of v. The greedy path is [(0, 1), (0, 1), (0, 1), (0, 1), (0, 1)]. The first two operations are the two scalar products [O(d)] and the third operation is a sum over an outer product [O(d^2)].

Ah sorry yes I see. At least personally these trivial cases of disconnected contractions haven't been on my radar which is maybe also why they aren't handled sensibly in various places. Combined with #99 it might be time to introduce some universal preprocessing of all the trivial contractions like these + hadamard products etc.

---

Here are the results, which look good! I think the apparent improvement over previous reuslts is mostly due to running with the full set of contractions as the others which I had not checked (sorry):

oe-dp-times

oe-dp-flops

Certainly this method seems to outperform ``branch-all``. I've also included ``random-greedy-128`` for reference as I think it might be nice to have an ``'auto-hq'`` mode that targets a longer solving time first using ``'dp'`` before switching over to maybe ``'random-greedy'``.

---

Just out of interest, can the ``dp`` approach be truncated and so made anytime? I.e. is there way to sort the search such that likely contractions are checked first or does it have to be exhaustive? Otherwise I'm thinking the ``BranchBound`` path finders might be refitted to fill that niche. In some brief tests ``branch-all`` performs very nicely if you just make it timeout rather than checking every solution. [No worries if you don't want to think about these things :)].

mrader1248:
Thank you very much for rerunning the benchmarks. It's very nice to see that `dp` is speedup-wise on par with the `optimal` approach, but can be used for even larger contractions.

Unfortunately, I don't see a trivial way to run a truncated version of `dp`. The `dp` optimiser simply finds the optimal contractions for all sets of n (connected) tensors by combining optimal contractions of all disjoint sets of m and n - m (connected) tensors. So I'm not sure if any other optimiser could use these intermediate results.

I implemented two more small code optimisations that should make `dp` run ~20-25 % faster. Before doing anything, my code converts all the n occurring indices to integers 0, 1, ..., n-1. I guess this could also be used to speed up other optimisers (e.g. operations on sets of integers are faster in CPython than on sets of strings). Another optimisation was to use bitmaps (integers) instead of tensor sets. Furthermore, FYI: I observed that ~40 % of `dp` runtime is spent in `compute_size_by_dict`.

dgasmith:
Not sure I have seen a way that speed up `dp` algorithms either.

The bitwise operations are interesting, since the `dp` algorithm can only handle pretty tame numbers of total indices this is likely fine for most cases. We might want to think if there are bounds in the number of supported input tensors and indices.

`compute_size_by_dict` is a good candidate to cythonize first. If we remap the code base to store sizes as an array and indices become integers that map that array we could make several items more efficient.

mrader1248:
As far as I understand it, this approach should be correct for an arbitrarily large number of inputs, as python automatically switches to some kind of big integer type. However, when using this big integer type, it could slow down these bitwise operations (which might still be faster than using the set classes).

Yes, to cythonize `compute_size_by_dict` might be a good idea, although this is beyond the scope of this pull request. However, keep in mind that I'm now calling this function with an array (=dict mapping integer indices to sizes) instead of a dict.

jcmgray:
Another update to the benchmarks then! By the way I've updated the notebook [here](https://gist.github.com/jcmgray/702a70c188b20e74506e635da2033ab0) that can generate these plots.

oe-times-graph

oe-flops-graph

Have you got any other planned updates @mrader1248? I'm thinking ``'dp'`` should probably replace ``'branch-all'`` in the ``'auto'`` path for ``len(inputs)=[5, 6]``, and at some point it would be good to add the dynamic programming approach to the proper documents, but maybe these could be done in a later, separate PR if you are keen to get this in?

mrader1248:
No, I'm fine now with my code. So I'm not planning to change anything else at the moment. I feel honoured that you want to put `dp` immediately into the `auto`-chain, but I'm not sure whether this should be accomplished with my pull request. The choice which optimisers should be part of the `auto`-chain is completely yours.

What to you mean with 'the proper documents'?

jcmgray:
> What to you mean with 'the proper documents'?

``opt_einsum`` has documentation e.g. [on the path finders](https://optimized-einsum.readthedocs.io/en/latest/path_finding.html) that is built from the files in ``opt_einsum/docs/source``, see for example [the page on the branching path](https://github.com/dgasmith/opt_einsum/blob/master/docs/source/branching_path.rst). At some point it would be good to include information on this new path.

> The choice which optimisers should be part of the auto-chain is completely yours.

@dgasmith what do you think? It seems like a fairly natural upgrade to me.

dgasmith:
I agree on replacing the branching algorithm here with ‘dp’, the statistics look good. At some point we are going to have to do a white paper on these algorithms. We are collecting enough it could be interesting.

mrader1248:
I remembered that a good upper bound for the contraction cost could give quite some speedup, so I added an optional parameter `cost_limit` for `dp` to reject contractions (already of subsets of the tensors) that exceed this limit. Then I also realised that for a larger number of inputs, I could use an estimate from `random-greedy-128` to speed up `dp`.

jcmgray:
Sounds like a good enhancement. Rather than adding a new cost function how possible would it be to make the cost the same (if it is not currently) as the flops estimate as used by the other contractors? It would be also nice to add ``cost_limit`` to the constructor of ``DynamicProgrammingOptimizer``. That would allow this kind of thing:

```python
rg_opt = oe.RandomGreedy()
oe.contract_path(eq, *arrays, optimize=rg_opt)

dp_opt = oe.DynamicProgrammingOptimizer(cost_limit=rg_opt.best['flops'])
oe.contract_path(eq, *arrays, optimize=dp_opt)
```
And does running just a simple `greedy` cost_limit for all sizes help?

mrader1248:
I'm not sure, how this flops estimate is determined, but it seems to be larger than the one I'm using (I want to have it as small as possible, but still large enough).

You are right, I moved `cost_limit` to the constructor.

I guess this should help quite a bit. Just take this example of 20 inputs (5x4 tensors connected on a square lattice):
```
inputs = [
    {0, 15}, {0, 1, 16}, {1, 2, 17}, {2, 18},
    {3, 15, 19}, {3, 4, 16, 20}, {4, 5, 17, 21}, {5, 18, 22},
    {6, 19, 23}, {6, 7, 20, 24}, {7, 8, 21, 25}, {8, 22, 26},
    {9, 23, 27}, {9, 10, 24, 28}, {10, 11, 25, 29}, {11, 26, 30},
    {12, 27}, {12, 13, 28}, {13, 14, 29}, {14, 30}
]
size_dict = {k: 2 for k in set.union(*inputs)}
```
`random-greedy-128` takes 157 ms ± 7.12 ms, `dp` without `cost_limit` ~ 100 s and `dp` with the `random-greedy-128` estimate required 563 ms ± 2.44 ms in my test.

jcmgray:
> I'm not sure, how this flops estimate is determined, but it seems to be larger than the one I'm using (I want to have it as small as possible, but still large enough).

Yes I mean ideally ``'dp'`` would use this same flop costing internally as well, based on calling ``helpers.flop_count``, then *any* cost calculated by another path would be a valid sieve for this.

mrader1248:
Oh, now I see how you meant it. Of course, this I could do. However, I don't quite understand the `flop_count` function. First of all, executing the examples from the docstring yield other results than stated in the docstring. Secondly, what is this `inner` parameter for?

jcmgray:
Yes I was confused a bit about [``flop_count`` and inner here](https://github.com/dgasmith/opt_einsum/pull/73#issuecomment-434897041), but as @dgasmith pointed out, it's because if there is an inner product, each element involves both a multiplication *and* a summation (subtlety ... unless the cpu instructions are executed with FMA support).

Regarding the docstring examples, looks like they are out-of-date and should be corrected.

At some point a possibility would be to have the current advanced flop estimate changeable so as to allow for instance the cost you are currently using instead. I think these are essentially the same (often just differ by a factor of 2) as ``'dp'`` currently finds the same paths as ``'optimal'``.

mrader1248:
If I just contract two tensors, `inner` has to be true if it is a scalar product, i.e. if there are no more indices left after performing this contraction? In the `dp` code I'm using now:
```
cost = cost1 + cost2 + helpers.flop_count(i1_union_i2, len(i) == 0, 2, size_dict)
```
The ugly thing about it: I have to compute `i` now before estimating the cost, which makes it more expensive..

But what's anyway quite strange: To get an upper bound, I use
```
rg = path_random.RandomGreedy(max_repeats=128)
p = rg(inputs, output, size_dict, memory_limit)
cost_limit = 1.1 * 0.5 * rg.best["flops"]
```
I can halve the flops from `random-greedy-128` and still get all the tests correct..

Also, I redefined the function `path_cost` I introduced for cross checking:
```
def path_cost(inputs, output, size_dict, path):
    c = 0
    for p in path:
        inputs_to_contract = [i for j, i in enumerate(inputs) if j in p]
        i_inputs_to_contract = set.union(*inputs_to_contract)
        inputs = [i for j, i in enumerate(inputs) if j not in p]
        i_contract = i_inputs_to_contract - output
        if len(inputs) > 0:
            i_contract -= set.union(*inputs)
        new_input = i_inputs_to_contract - i_contract
        c += flop_count(i_inputs_to_contract, len(new_input) == 0, len(inputs_to_contract), size_dict)
        inputs += [new_input]
    return c
```
If I use this function to recompute the cost of the `random-greedy-128` path, it's only half as much as `rg.best["flops"]`. So, is this function wrong or is there a bug in the flops estimate of `random-greedy-128`?

jcmgray:
Yes I think you need to tweak which indices are supplied. Here's a version that works for flops:

```python
def path_cost_flops(inputs, output, size_dict, path):
    inputs = list(inputs)
    c = 0
    for p in path:
        inputs_to_contract = [inputs.pop(i) for i in sorted(p, reverse=True)]
    
        either = set.union(*inputs_to_contract)
        shared = set.intersection(*inputs_to_contract)
        keep = set.union(output, *inputs)
    
        c += helpers.flop_count(either, shared - keep, len(p), size_dict)
        inputs.append(either & keep)
        
    return c
```
If it is significantly more expensive to keep track of flops like this then maybe sticking to the original cost is the best thing to do. I might actually run some timings to see how well each estimated cost correlates with real contraction time.

jcmgray:
Getting slightly off track here but this is quite interesting. If you generate random expressions, and compute the full flop cost (as currently done) and the 'number of operations' cost, then compared to the actual time to perform the computation, the slightly simpler 'number of operations' does seem like a slightly better predictor:

First is histogram of ``log10(flop_cost / time)``.
Second is histogram of ``log10(2 * num_ops / time)``.
Both are coloured by number of output indices.

visualization(4)

Both predict the actual time will be about 1e-8 times the cost, and they match when ``n_out == 0``. The second is very slightly tighter but also has no dependency on ``n_out``.

All of which is to say, it would not necessarily be a bad thing to simplify the FLOP counting across the board.

mrader1248:
I do not have high precision measurements, but `dp` is certainly 30-40 % slower when I have to compute the remaining indices of a contraction beforehand. So from this point of view it might really make sense to stick to the current estimators.

Could you be so kind as to rerun the `dp` benchmarks? In contrast to the previous changes I guess these latest changes should really affect the optimisation times.

jcmgray:
I'll finish running the updated benchmarks soon - as @mrader1248 says, implementing the cost limit with a greedy call should make a big difference! But otherwise agree with leaving the other stuff to different issues.

mrader1248:
I made a last modification: Instead of running random-greedy in the beginning, I am now using an iteratively increasing cost cap. For the example of 5x4 tensors on a square lattice this brings me down from 563 ms to 420 ms and I can do 6x5 tensors in less than 30 s. The problem with the cost estimate from random-greedy is that for a larger number of inputs it becomes too bad to really speed up dp. Of course, also with the iterative cost cap dp scales exponentially with the number of inputs, but the overall performance is on the one hand side a bit better and on the other hand side the dependency on random-greedy is removed.

@dgasmith, I also think -- if you are okay with the latest version of my pull request -- that it is time to merge. Regarding the points you mentioned above: I think that these four points should be tackled in new issues/pull requests.

jcmgray:
Happy to merge if you want to get this in - not trying to delay! 

However I think the cost cap stuff does merit a bit more thinking about maybe in a later PR as there likely gains to be had. For any planar graphs (such as lattices) the random-greedy approach generally finds the optimal path, indeed, I find for both 4x5 and 5x6 lattices it finds the exact same path cost as ``'dp'``. So it seems surprising it wasn't working as a good cost cap. Out of interest, why did you opt for ``int(1.1 * cost_limit)``, surely any valid path cost found is a valid limit?

Here is a graph of timings including the newest method (``'dp-costcap``), and I ran the greedy labelled ones with that 1.1 scaling removed:

oe-cost-time

It might be worth in the end having options for:
* using no costlimit
* using the iterative method
* supplying a str or PathOptimizer to precompute one
* using some automatically selected number of greedy runs

dgasmith:
Tiers of greedy runs could be the name of the day:
```python
ngreedy = 0
if nterms > 7:
   ngreedy = (terms - 7) ** 2
ngreedy = max(ngreedy, 128)
```

Though simply `greedy-1` seems to do a reasonable job across the board.

Not trying to push this through too quickly. Just wanted to rule a few items that are in and out of scope of this PR so that it doesn't go on forever :)

mrader1248:
Maybe it's really the best to just merge this for the moment to have dp in opt_einsum and then start  thinking about possibilities of using combinations of dp and random-greedy. I would go for the version with the cost cap as it's completely self-contained and the performance is not bad.

@jcmgray, I used the factor `1.1` as I wasn't entirely sure whether there could be a small +epsilon in the cost. But of course, if this isn't the case one can omit this factor.

@dgasmith, I already realised that with an increasing number of inputs also random-greedy-X with larger X should be used. However, I have the bad feeling that the optimal X might also scale exponentially with the number of inputs.

Do you want to have any further changes in this PR or are you okay with merging it?

dgasmith:
Well, not sure if we are really after *optimal* X, but just something reasonable at first.

Looking over again, now. But overall LGTM.

@jcmgray Are you ok with merging?

opt_einsum/paths.py
```diff
@@ -652,6 +655,250 @@ def greedy(inputs, output, size_dict, memory_limit=None, choose_fn=None, cost_fn
     return ssa_to_linear(ssa_path)
 
 
+def _tree_to_sequence(c):
+    # converts a contraction tree to a contraction sequence, e.g.
```
dgasmith: Can we convert these to `"""` doc strings with a small example?

opt_einsum/paths.py
```diff
@@ -652,6 +655,250 @@ def greedy(inputs, output, size_dict, memory_limit=None, choose_fn=None, cost_fn
...
+
+        for i in [i for i in j if type(i) != int]:
+            s[0] += (len(t) + len(c),)
+            c.append(i)
+
+    return s
+
+
+def _find_disconnected_subgraphs(inputs, output):
+    # finds disconnected subgraphs in the given list of inputs and returns
```
dgasmith: Looks like for all of these functions. The `e.g. inputs` section is good, just please do convert to a canonical NumPy-style doc string.

opt_einsum/paths.py
```diff
@@ -652,6 +655,250 @@ def greedy(inputs, output, size_dict, memory_limit=None, choose_fn=None, cost_fn
...
+    # e.g. 0b1001011 --> [0, 1, 3, 6]
+    j = 0
+    while s != 0:
+        if s & 1 != 0:
+            yield j
+        s >>= 1
+        j += 1
+
+
+class DynamicProgrammingOptimizer(PathOptimizer):
```
dgasmith: Can we add a docstring here similar to `BranchBound`?

opt_einsum/paths.py
```diff
@@ -652,6 +655,250 @@ def greedy(inputs, output, size_dict, memory_limit=None, choose_fn=None, cost_fn
...
+            # output index dimensions as initial cost_cap
+            cost_cap = helpers.compute_size_by_dict(
+                set.union(*(inputs[j] for j in _bitmapset_indices(g))) & output,
+                size_dict
+            )
+
+            while len(x[-1]) == 0:
+                for n in range(2, len(x[1]) + 1):
+                    xn = x[n]
+                    for m in range(1, n // 2 + 1):  # try to combine solutions from x[m] and x[n-m]
```
dgasmith: Can we move the comments to above the code line and place a single line break above?

jcmgray:
Yes happy to merge as is - I've updated that latest graph with ``n=20`` and the various methods look pretty close in the limit. Presumably those last few review comments should be addressed first, then yes all looks good to me!

mrader1248:
> Well, not sure if we are really after _optimal_ X, but just something reasonable at first.

The big advantage of the cost capping strategy is that we don't need to worry about this X.

@dgasmith, I incorporated your suggestions regarding documentation. From my point of view we are now ready to merge.

PR #106, opened 2019-09-09
added dp documentation

mrader1248:
## Description
Added documentation for the dynamic programming (DP) path optimizer.


jcmgray:
Thanks for adding these! All looks good to me and I think I finally understand the internals. The only remaining thing to do is to add a reference in the [index doc](https://github.com/dgasmith/opt_einsum/blob/master/docs/source/index.rst), then it will appear on readthedocs etc.

jcmgray:
Thanks again for adding this optimizer - will be very useful. For future reference, does 'asymptotically' optimal refer to increasing bond sizes?

mrader1248:
Thank you as well for your support.

Exactly, 'asymptotically optimal' refers to increasing dimensions. Consider for example the equation `"a,b,abc->c"` (and let's call the dimensions `da`, `db`, and `dc`. The computational complexity of this would be `O(da*db*dc + da*db)`, `O(da*db*dc + da*dc)`, or `O(da*db*dc + db*dc)`, depending on which contraction you perform first. For the one with `+ da*db`  one has to compute an other product first, which will never be done by the DP approach. So, if for example `da*db < da*dc < db*dc`, it would be better to compute the outer product first, but asymptotically all these three contractions paths are equal.

PR #109, opened 2019-11-08
add mars support via alias

jcmgray:
## Description
Add support for [mars](https://github.com/mars-project/mars) (just needs a simple alias), a library a bit similar to `dask`.

## Status
- [x] Ready to go


jcmgray:
An alternative to adding aliases like this (which is admittedly pretty trivial...) might be to have a `register_alias` function for other libraries to use.

PR #111, opened 2019-11-27
PyPI Stats Badge

dgasmith:
## Description
We seem to have hit a new level of 1.5M+ downloads/month which is quite something. I thought it might be nice to add a badge to show this off.

As a side note does anyone know which package we went into early October which caused the spike? I think the underscore in the package name is preventing GitHub from correctly finding dependencies and I am unsure how to track it through PyPI or Conda without construct the full dependency tree locally. 

At this scale it would be good to be in contact with all downstream packages like we are with Pyro.

## Status
- [x] Ready to go


jcmgray:
> 1.5M+ downloads/month

Wow that is quite something! Look's like [tensorflow](https://github.com/tensorflow/tensorflow/commit/8b01b9c5090fc3ceb72a83d66123d0050d11e705) could be the culprit?

Issue #118, opened 2019-12-02
optimize='dp' gets confused by singleton factors 


yaroslavvb:
This hangs with 100% utilization. 
```
import opt_einsum as oe
einsum_string = 'nlp,nlq->n'
views = oe.helpers.build_views(einsum_string, {'n': 30, 'i': 100, 'l': 100, 'p': 2, 'q': 2, 'z': 1})
path, path_info = oe.contract_path(einsum_string, *views,optimize='dp')
print(path_info)
```
Changing optimizer or increasing size of `z` factor works

jcmgray:
Ah this is because of the costcapping strategy. I'll fix this in the upcoming tweaks to 'dp'.

jcmgray:
This should be fixed by #119.

Issue #116, opened 2019-11-30
Getting "not enough values to unpack" with optimize=dp


yaroslavvb:
The following breaks with optimize='dp' but works with default optimizer
```
import opt_einsum as oe 
einsum_string = 'a,bcd,efg->'
views = oe.helpers.build_views(einsum_string, {c: 2 for c in einsum_string.replace(',->','')})
path, path_info = oe.contract_path(einsum_string, *views, optimize='dp')
print(path_info)
```
produces
```
/usr/local/lib/python3.6/dist-packages/opt_einsum/contract.py in contract_path(*operands, **kwargs)
    270     else:
    271         path_optimizer = paths.get_path_fn(path_type)
--> 272         path = path_optimizer(input_sets, output_set, dimension_dict, memory_arg)
    273 
    274     cost_list = []

/usr/local/lib/python3.6/dist-packages/opt_einsum/paths.py in dynamic_programming(inputs, output, size_dict, memory_limit)
    971 def dynamic_programming(inputs, output, size_dict, memory_limit=None):
    972     optimizer = DynamicProgrammingOptimizer()
--> 973     return optimizer(inputs, output, size_dict, memory_limit)
    974 
    975 

/usr/local/lib/python3.6/dist-packages/opt_einsum/paths.py in __call__(self, inputs, output, size_dict, memory_limit)
    872         inputs, inputs_contractions = zip(*[
    873             (i - i_single, j if i.isdisjoint(i_single) else (j,))
--> 874             for j, i in enumerate(inputs)
    875             if len(i - i_single) > 0
    876         ])

ValueError: not enough values to unpack (expected 2, got 0)
```

jcmgray:
I'm making a few small updates to ``'dp'`` and will include a fix for this this shortly.

jcmgray:
This should be fixed by #119.

PR #119, opened 2019-12-02
updates to 'dp' optimizer

### LINKED ISSUE #99: `Optimal` path not optimal?

jcmgray:
## Description
These updates the ``'dp'`` optimizer in a few ways:
- fixes some edge case bugs (#116, #118) @yaroslavvb
- allows it to be customized in a few ways (without any impact on performance in the default case):
    * whether to target minimum flop count or minimize size
    * whether to search through outer products
    * how to set the ``cost_cap`` iterate strategy (on, off or manual)
- a few tweaks to make the overhead lower for small contractions particularly
- renames ``DynamicProgrammingOptimizer->DynamicProgramming`` and add it to the top namespace

I think with these, it would make sense to make the ``'optimal'`` algorithm point at ``'dp'`` since with the outer product search it should now find all the same contractions, and with the ``cost_cap`` turned off it's as fast for small contractions as well. This would fix #99. Thoughts @dgasmith? Also cc @mrader1248.

## Todos
  - [x] document customizable `DynamicProgramming`
  - [ ] one could identify now 'standard tensor network' contractions (i.e. every index appears exactly twice) at the beginning from the index count, and possibly switch to a faster method for computing resulting indices (simply the symmetric difference of inputs)
  - [ ] use a version of ``'dp'`` for ``'optimal'``?
  - [ ] have an ``'auto'``value for the `cost_cap` that simply turns it off til the contraction is quite big

## Status
- [ ] Ready to go

### DIFF
--- a/docs/source/dp_path.rst
+++ b/docs/source/dp_path.rst
@@ -23,9 +23,9 @@ complexity of finding the contraction path is drastically reduced: If the
 subgraphs consist of ``n1``, ``n2``, ... inputs, the computational complexity
 is reduced from ``O(exp(n1 + n2 + ...))`` to ``O(exp(n1) + exp(n2) + ...)``.
 
-The DP approach will only perform pair contractions and will never compute
-intermediate outer products as in reference [1] it is shown that this always
-results in an asymptotically optimal contraction path.
+The DP approach will only perform pair contractions and by default will never
+compute intermediate outer products as in reference [1] it is shown that this
+always results in an asymptotically optimal contraction path.
 
 A major optimization for DP is the cost capping strategy: The DP optimization
 only memorizes contractions for a subset of inputs, if the total cost for this
@@ -41,5 +41,30 @@ contraction graphs and is guaranteed to find an asymptotically optimal
 contraction path. For this reason it is the most frequently used contraction
 path optimizer in the field of tensor network states.
 
+More specifically, the search is performed over connected subgraphs, which, for
+example, planar and tree-like graphs have far fewer of. As a rough guide, if
+the graph is planar, expressions with many tens of tensors are tractable,
+whereas if the graph is tree-like, expressions with many hundreds of tensors
+are tractable.
+
 
 [1] Robert N. C. Pfeifer, Jutho Haegeman, and Frank Verstraete Phys. Rev. E 90, 033315 (2014). https://arxiv.org/abs/1304.6112
+
+
+Customizing the Dynamic Programming Path
+----------------------------------------
+
+The default ``optimize='dp'`` approach has sensible defaults but can be
+customized with the :class:`~opt_einsum.paths.DynamicProgramming` object.
+
+.. code:: python
+
+    import opt_einsum as oe
+
+    optimizer = oe.DynamicProgramming(
+        minimize='size',    # optimize for size (rather than FLOPs)
+        search_outer=True,  # search through outer products as well
+        cost_cap=False,     # don't use cost-capping strategy
+    )
+
+    oe.contract(eq, *arrays, optimize=optimizer)
--- a/docs/source/dp_path.rst
+++ b/docs/source/dp_path.rst
@@ -68,3 +68,8 @@ customized with the :class:`~opt_einsum.paths.DynamicProgramming` object.
     )
 
     oe.contract(eq, *arrays, optimize=optimizer)
+
+.. warning::
+
+    Note that searching outer products will most likely drastically slow down
+    the optimizer on all but the smallest examples.
--- a/docs/source/dp_path.rst
+++ b/docs/source/dp_path.rst
@@ -62,7 +62,7 @@ customized with the :class:`~opt_einsum.paths.DynamicProgramming` object.
     import opt_einsum as oe
 
     optimizer = oe.DynamicProgramming(
-        minimize='size',    # optimize for size (rather than FLOPs)
+        minimize='size',    # optimize for largest intermediate tensor size
         search_outer=True,  # search through outer products as well
         cost_cap=False,     # don't use cost-capping strategy
     )
--- a/opt_einsum/paths.py
+++ b/opt_einsum/paths.py
@@ -772,25 +772,26 @@ def _bitmap_select(s, seq):
     return (x for x, b in zip(seq, bin(s)[:1:-1]) if b == '1')
 
 
+def _dp_calc_legs(g, all_tensors, s, inputs, i1_cut_i2_wo_output, i1_union_i2):
+    # set of remaining tensors (=g-s)
+    r = g & (all_tensors ^ s)
+    # indices of remaining indices:
+    if r:
+        i_r = set.union(*_bitmap_select(r, inputs))
+    else:
+        i_r = set()
+    # contraction indices:
+    i_contract = i1_cut_i2_wo_output - i_r
+    return i1_union_i2 - i_contract
+
+
 def _dp_compare_flops(cost1, cost2, i1_union_i2, size_dict, cost_cap, s1, s2, xn, g,
                       all_tensors, inputs, i1_cut_i2_wo_output, memory_limit, cntrct1, cntrct2):
     cost = cost1 + cost2 + helpers.compute_size_by_dict(i1_union_i2, size_dict)
     if cost <= cost_cap:
         s = s1 | s2
         if s not in xn or cost < xn[s][1]:
-            # set of remaining tensors (=g-s)
-            r = g & (all_tensors ^ s)
-
-            # indices of remaining indices:
-            if r:
-                i_r = set.union(*_bitmap_select(r, inputs))
-            else:
-                i_r = set()
-
-            # contraction indices:
-            i_cntrct = i1_cut_i2_wo_output - i_r
-
-            i = i1_union_i2 - i_cntrct
+            i = _dp_calc_legs(g, all_tensors, s, inputs, i1_cut_i2_wo_output, i1_union_i2)
             mem = helpers.compute_size_by_dict(i, size_dict)
             if memory_limit is None or mem <= memory_limit:
                 xn[s] = (i, cost, (cntrct1, cntrct2))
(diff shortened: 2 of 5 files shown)


opt_einsum/paths.py
```diff
@@ -914,41 +997,27 @@ def __call__(self, inputs, output, size_dict, memory_limit=None):
                     # try to combine solutions from x[m] and x[n-m]
                     for m in range(1, n // 2 + 1):
                         for s1, (i1, cost1, cntrct1) in x[m].items():
-                            for s2, (i2, cost2, cntrct2) in x[n-m].items():
+                            for s2, (i2, cost2, cntrct2) in x[n - m].items():
 
                                 # only if s1 and s2 are disjoint
-                                if s1 & s2 == 0:
+                                if not s1 & s2:
```
dgasmith: Can we collapse this and the below if to a single line to help with the indention?

opt_einsum/tests/test_paths.py
```diff
@@ -197,6 +197,59 @@ def test_greedy_edge_cases():
     assert check_path(path, [(0, 1), (0, 2), (0, 1)])
 
 
+def test_dp_edge_cases_dimension_1():
```
dgasmith: Awesome, thanks for the extra tests!

opt_einsum/paths.py
```diff
@@ -761,38 +761,77 @@ def _find_disconnected_subgraphs(inputs, output):
...
-    >>> type(_bitmapset_indices(0b1001011))
-    generator
+            # indices of remaining indices:
+            if r:
+                i_r = set.union(*_bitmap_select(r, inputs))
+            else:
+                i_r = set()
+
+            # contraction indices:
+            i_cntrct = i1_cut_i2_wo_output - i_r
```
dgasmith: Can we write out `cntrct` (is it contract?), will help with readability.

opt_einsum/paths.py
jcmgray: Ah yes nice.

opt_einsum/paths.py
jcmgray: I think it is (I didn't write these variable names originally), but yes can update.

docs/source/dp_path.rst
```diff
@@ -41,5 +41,30 @@ contraction graphs and is guaranteed to find an asymptotically optimal
...
+
+The default ``optimize='dp'`` approach has sensible defaults but can be
+customized with the :class:`~opt_einsum.paths.DynamicProgramming` object.
+
+.. code:: python
+
+    import opt_einsum as oe
+
+    optimizer = oe.DynamicProgramming(
+        minimize='size',    # optimize for size (rather than FLOPs)
```
dgasmith: You may want to define `size` here.

opt_einsum/paths.py
```diff
@@ -761,38 +761,66 @@ def _find_disconnected_subgraphs(inputs, output):
...
+    return (x for x, b in zip(seq, bin(s)[:1:-1]) if b == '1')
 
-    Examples
-    --------
-    >>> type(_bitmapset_indices(0b1001011))
-    generator
 
-    >>> list(_bitmapset_indices(0b1001011))
-    [0, 1, 3, 6]
+def _dp_calc_legs(g, all_tensors, s, inputs, i1_cut_i2_wo_output, i1_union_i2):
```
dgasmith: Can we add short doc strings for these two functions.

opt_einsum/paths.py
```diff
@@ -841,53 +915,57 @@ def __call__(self, inputs, output, size_dict, memory_limit=None):
...
+        inputs, inputs_done, inputs_contractions = [], [], []
+        for j, i in enum_inputs:
+            i_reduced = i - i_single
+            if not i_reduced:
+                # input reduced to scalar already - remove
+                inputs_done.append((j,))
+            else:
+                # if the input has any index reductions, add single contraction
+                inputs.append(i_reduced)
+                inputs_contractions.append((j,) if i_reduced != i else j)
```
mrader1248: I see that my original code did not work if there was nothing more to optimise at this point. However, I find this reassignment of `inputs` in combination with the loop rather confusing.. I guess the original code could be fixed, if  the `zip` would only be unpacked if the list constructed inside the `zip` is not empty.

opt_einsum/paths.py
mrader1248: Yes, it stands for contract. Don't know why I named it that way.. Maybe some line length issue..

opt_einsum/tests/test_paths.py
mrader1248: Good idea to add these tests.

mrader1248:
Can you give me an example, where it is beneficial to construct outer products?

mrader1248:
Sorry, it took me a while to have a look at this. Overall, LGTM. And especially thanks for fixing this stupid smallest dimension 1 bug.

jcmgray:
> Can you give me an example, where it is beneficial to construct outer products?

I've put this example in the tests (taken from you in another thread I think!):
```python
def test_custom_dp_can_optimize_for_outer_products():
    eq = "a,b,abc->c"

    da, db, dc = 2, 2, 3
    shapes = [(da,), (db,), (da, db, dc)]

    opt1 = oe.DynamicProgramming(search_outer=False)
    opt2 = oe.DynamicProgramming(search_outer=True)

    info1 = oe.contract_path(eq, *shapes, shapes=True, optimize=opt1)[1]
    info2 = oe.contract_path(eq, *shapes, shapes=True, optimize=opt2)[1]

    assert info2.opt_cost < info1.opt_cost

```
Of course in practice I imagine it's almost never useful. However I think it would make sense for ``'dp'`` to replace the optimal implementation, and if that happens then the *option* to be guaranteed optimal seems beneficial for testing edge cases etc. Plus for v small sizes there's no penalty hit.

@dgasmith do you have any thoughts on pointing ``'optimal'`` to ``'dp'``? Maybe in a later PR.

opt_einsum/paths.py
jcmgray: Yes this seemed liked the simplest way to solve the 'nothing left to do' bug and also not iterate and filter the inputs twice in the same way. But certainly open to suggestions for making it more readable. Maybe just renaming the new ``inputs ->  inputs_parsed`` or something?

jcmgray:
@dgasmith Also happy to have a go a adding more docs (e.g. each variable etc) if you think necessary.

mrader1248:
@jcmgray I see, I thought you found an example where constructing outer products first leads to a better solution even in the asymptotic limit. Maybe it would be good to mention in the docs that allowing for intermediate outer products can lead to horrible optimisation times?

opt_einsum/paths.py
mrader1248: To be honest, for me list comprehensions are much more expressive compared to list constructions using loops. However, at least the variable `inputs` should be renamed.

jcmgray: Of course generally I agree! But when for loops can do the same thing more efficiently... I think they are the correct tool to use. I've renamed the parsed ``inputs`` to ``inputs_contract``.

jcmgray:
> @jcmgray I see, I thought you found an example where constructing outer products first leads to a better solution even in the asymptotic limit

Well in the asymptotic limit where ``da`` and ``db`` are fixed but ``dc`` get's larger, indeed the outer product is still necessary here.

> Maybe it would be good to mention in the docs that allowing for intermediate outer products can lead to horrible optimisation times? 

I've mentioned this in the docstring for the ``search_outer`` option, but yes might be worth adding another warning to the main docs as well.

jcmgray:
OK, so I've added a warning to the ``'dp'`` docs about turning on the outer product search, and factored out the single term parsing of inputs. Unless there are more comments I'll merge in a bit.

PR #120, opened 2019-12-21
Publish sdist and bdist wheel

groodt:
## Description
The benefits of wheels are well documented. See: https://pythonwheels.com/
This package is pure Python and publishing it as both source and as a wheel is simple.


groodt:
Thanks for the review @dgasmith

PR #124, opened 2020-02-02
adding Python 3.8 testing in the matrix

stonebig:
## Description
Adding Python 3.8 in the Matrix of test

## Status
- [ ] Ready to go

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -13,6 +13,7 @@ matrix:
     - python: 3.6
     - python: 3.7
     - python: 3.7
+    - python: 3.8
 
 matrix:
   include:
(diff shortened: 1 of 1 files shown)


.travis.yml
```diff
@@ -42,6 +43,12 @@ matrix:
       - PYTHON_VER=3.7
       - ENVIRON=FULL
 
+
+    - env:
+      - NAME="FULL 3.8"
+      - PYTHON_VER=3.8
+      - ENVIRON=FULL
```
dgasmith: This should be `MINIMAL`.

.travis.yml
```diff
@@ -47,7 +47,7 @@ matrix:
     - env:
       - NAME="FULL 3.8"
```
dgasmith: We would need to change the name here too; however the FULL test worked so we can switch it back. Please do merge #122 and #124 together. Either PR is fine.

stonebig: I don't know how to merge two PR with github web interface. please redo your own PR, if you want this. Glad to know Python-3.8 will work

dgasmith: Closing the previous one and recommitting to this branch will work as well. I don't think you can merge through the GUI either.

Issue #126, opened 2020-02-28
modify python_requires in setup.py to require python 3.6+


cdepillabout:
The `setup.py` file sets `python_requires='>=3.5'`:

https://github.com/dgasmith/opt_einsum/blob/4323cfce5f415dc425143f216171cefa56d5d429/setup.py#L25

However, `opt_einsum` is using `random.choices`, which is only available on `>=python-3.6`:

https://github.com/dgasmith/opt_einsum/blob/4323cfce5f415dc425143f216171cefa56d5d429/opt_einsum/path_random.py#L263

`random.choices` api docs: https://docs.python.org/3/library/random.html#random.choices

----------------------------------

Also, as far as I can tell, travis is attempting to test with `python-3.5`:

https://github.com/dgasmith/opt_einsum/blob/4323cfce5f415dc425143f216171cefa56d5d429/.travis.yml#L12

https://github.com/dgasmith/opt_einsum/blob/4323cfce5f415dc425143f216171cefa56d5d429/.travis.yml#L21-L24

But these tests are actually running with `python-3.6`:

- https://travis-ci.org/dgasmith/opt_einsum/jobs/645102401#L182
- https://travis-ci.org/dgasmith/opt_einsum/jobs/645102401#L488
- https://travis-ci.org/dgasmith/opt_einsum/jobs/645102401#L550

dgasmith:
Thanks for the report, this is indeed a mistake and a failure with the current CI setup. For fixing this particular issue I believe we can replace with with `np.random.choice`. I think this is 1-1, @jcmgray can you comment?

We are due for a release so we should be able to patch this up and get a release out in the next few days.

jcmgray:
Yep I can fix this shortly - the reason to use ``random.choices`` is speed, so if there are no objections I'll try and import that version first then resort to numpy.

PR #128, opened 2020-03-01
CI: Ensures correct python build versions

dgasmith:
## Description
Fixes CI to build with the declared Python version. Ensures #126 does not happen again.

## Status
- [x] Ready to go


dgasmith:
@jcmgray Can you double check the [Py35](https://github.com/dgasmith/opt_einsum/pull/128/commits/3a083bdd3ac9d3dcf21857e8246ba0a0e433cbbf) edits?

dgasmith:
Yes, without OrderedDict `dp` gives non-deterministic results.

A small test shows initialization time is about 3x larger, but averages to microseconds for our use case. Lookup time is equivalent and iteration time is about 1.3x larger. Average over several tests show the `dp` takes a 2-4% performance hit. Not the best, but keeps things simple.

jcmgray:
Cool, thanks for the numbers! I agree that seems reasonable.

dgasmith:
Great! I will go ahead with a release in this case.

PR #134, opened 2020-04-04
Fix some examples

kngwyu:
## Description
Thank you for this great work.
This PR fixes the example of the top page to work by just copy-pasting the code.

And I also added `import opt_einsum as oe` just before the first time it appers.

### DIFF
--- a/docs/source/index.rst
+++ b/docs/source/index.rst
@@ -107,7 +107,7 @@ Details about the optimized contraction order can be explored:
 
 .. code:: python
 
-    >>> from opt_einsum import contract_path
+    >>> import opt_einsum as oe
 
     >>> path_info = oe.contract_path('pi,qj,ijkl,rk,sl->pqrs', C, C, I, C, C)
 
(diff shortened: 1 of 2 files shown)


docs/source/index.rst
```diff
@@ -109,7 +109,7 @@ Details about the optimized contraction order can be explored:
 
     >>> from opt_einsum import contract_path
 
-    >>> path_info = oe.contract_path('pi,qj,ijkl,rk,sl->pqrs', C, C, I, C, C)
+    >>> path_info = contract_path('pi,qj,ijkl,rk,sl->pqrs', C, C, I, C, C)
```
dgasmith, line 112: Thanks for catching this, it looks like we use `import opt_einsum as oe` much more than individual function imports in the docs. We need to do a general homogenization pass, but for this example at least can you change line 110 instead of 112?

kngwyu: I did, thank you for the review.

PR #135, opened 2020-04-15
Make JAX backend import lazy.

hawkinsp:
We observed the following surprising behavior: importing tensorflow also imports jax, if installed. This happens because tensorflow has an opt_einsum dependency that is imported eagerly, and the opt_einsum jax backend eagerly imports jax. To avoid this, make the jax import from opt_einsum lazy.

## Status
- [ * ] Ready to go

### DIFF
--- a/devtools/travis-ci/before_install.sh
+++ b/devtools/travis-ci/before_install.sh
@@ -13,7 +13,7 @@ else
     MINICONDA=Miniconda3-latest-Linux-x86_64.sh
 fi
 MINICONDA_HOME=$HOME/miniconda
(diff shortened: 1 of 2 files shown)


hawkinsp:
It looks like all CI builds are failing for reasons related to Conda, not the PR.

dgasmith:
The CI failure may be enough impetus to switch the repo over to GH actions.

hawkinsp:
Done. The CI builds passed.

A new release would be welcome, thanks!

dgasmith:
Thanks for the CI fix, I pull put out a patch momentarily.

dgasmith:
Release is up, conda may take a bit to build.

PR #141, opened 2020-05-12
stateful path optimizers check args + minor doc tweaks

### LINKED ISSUE #137: Possible bug in RandomGreedy path
I have found an interesting quirk of the RandomGreedy optimizer that may not have been intended. In my application, I need to repeat two contractions several times. I use the same optimizer to compute the paths in advance and cache the result. If I use RandomGreedy, then I get the wrong shape for the second contraction, which leads me to believe that there's some internal state of RandomGreedy that persists into computing the second path. Below is a MWE.

```python
import torch
import opt_einsum as oe

eqn1 = 'abc,bef,ahi,ehk,uvc,vwf,uxi,wxk->'
eqn2 = 'abd,beg,ahj,ehl,mndo,npgq,mrjs,prlt,uvo,vwq,uxs,wxt->'
shapes1 = [(1, 1, 2), (1, 1, 2), (1, 1, 2), (1, 1, 2), 
           (1, 1, 2), (1, 1, 2), (1, 1, 2), (1, 1, 2)] 
shapes2 = [(1, 1, 2), (1, 1, 2), (1, 1, 2), (1, 1, 2), (1, 1, 2, 2), (1, 1, 2, 2), 
           (1, 1, 2, 2), (1, 1, 2, 2), (1, 1, 2), (1, 1, 2), (1, 1, 2), (1, 1, 2)] 
tens1 = [torch.randn(*sh) for sh in shapes1]
tens2 = [torch.randn(*sh) for sh in shapes2]

for Opt in [oe.DynamicProgramming, oe.RandomGreedy]:
    opt = Opt(minimize='flops')
                                                                                                                                 
    expr1 = oe.contract_expression(eqn1, *[ten.shape for ten in tens1], optimize=opt)
    expr2 = oe.contract_expression(eqn2, *[ten.shape for ten in tens2], optimize=opt)

    print(expr1(*tens1))
    print(expr2(*tens2))
```

The output of this program, using the latest PyPI release, is:

```
tensor(-0.0187)
tensor(0.4303)
tensor(-0.0187)
tensor([[[[[-0.4418]],

          [[ 1.2737]]]]])
```

Lines 1 and 3 should match (and do), but so should 2 and 4 (and don't).

jcmgray:
## Description
Explicitly check that stateful path optimizers are called with the same arguments on repeat runs (should resolve #137), and error otherwise. I also updated the ``contract`` docstring with the newer ``optimize`` options, and renamed a (purely internal) variable ``dimension_dict -> size_dict`` for consistency with code elsewhere.

## Status
- [x] Ready to go

### DIFF
--- a/opt_einsum/contract.py
+++ b/opt_einsum/contract.py
@@ -214,7 +214,7 @@ def contract_path(*operands, **kwargs):
     indices = set(input_subscripts.replace(',', ''))
 
     # Get length of each unique dimension and ensure all dimensions are correct
-    dimension_dict = {}
+    size_dict = {}
     for tnum, term in enumerate(input_list):
         sh = input_shps[tnum]
 
@@ -224,18 +224,18 @@ def contract_path(*operands, **kwargs):
         for cnum, char in enumerate(term):
             dim = int(sh[cnum])
 
-            if char in dimension_dict:
+            if char in size_dict:
                 # For broadcasting cases we always want the largest dim size
-                if dimension_dict[char] == 1:
-                    dimension_dict[char] = dim
-                elif dim not in (1, dimension_dict[char]):
+                if size_dict[char] == 1:
+                    size_dict[char] = dim
+                elif dim not in (1, size_dict[char]):
                     raise ValueError("Size of label '{}' for operand {} ({}) does not match previous "
-                                     "terms ({}).".format(char, tnum, dimension_dict[char], dim))
+                                     "terms ({}).".format(char, tnum, size_dict[char], dim))
             else:
-                dimension_dict[char] = dim
+                size_dict[char] = dim
 
     # Compute size of each input array plus the output array
-    size_list = [helpers.compute_size_by_dict(term, dimension_dict) for term in input_list + [output_subscript]]
+    size_list = [helpers.compute_size_by_dict(term, size_dict) for term in input_list + [output_subscript]]
     memory_arg = _choose_memory_arg(memory_limit, size_list)
 
     num_ops = len(input_list)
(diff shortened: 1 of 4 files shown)


opt_einsum/paths.py
```diff
@@ -43,6 +43,18 @@ def __call__(self, inputs, output, size_dict, memory_limit=None):
 
     where ``path`` is a list of int-tuples specifiying a contraction order.
     """
+
+    def check_args_against_first_call(self, inputs, output, size_dict):
+        """Utility that stateful optimizers can use to ensure they are not
+        called with different contractions across separate runs.
+        """
+        args = (inputs, output, size_dict)
+        if not hasattr(self, '_first_call_args'):
```
dgasmith, line 52: Initializing this to `None` might be simpler.

```
if self._first_call_args is not None:
```

opt_einsum/paths.py
```diff
@@ -43,6 +43,18 @@ def __call__(self, inputs, output, size_dict, memory_limit=None):
 
     where ``path`` is a list of int-tuples specifiying a contraction order.
     """
+
+    def check_args_against_first_call(self, inputs, output, size_dict):
```
dgasmith, line 47: Can we make this private? `_check` looks like an internal helper function rather than user facing.

opt_einsum/contract.py
```diff
@@ -412,7 +423,7 @@ def contract(*operands, **kwargs):
 
         The default is None. Note that imposing a limit can make contractions
         exponentially slower to perform.
-    backend : str, optional (default: ``numpy``)
+    backend : str, optional (default: ``auto``)
```
dgasmith, line 426: Nice!

opt_einsum/paths.py
jcmgray: Yes sounds good.

opt_einsum/paths.py
jcmgray: It would be preferable, the only reason I haven't is that ``PathOptimizer`` itself doesn't have an ``__init__`` so there is no single place to initialize the ``_first_call_args``.

dgasmith: Good point, can you add a comment to this effect?

jcmgray: Have done, let me know if it needs more clarification!

PR #161, opened 2021-05-15
Python Type Annotations

dgasmith:
## Description
Bumps minimum python version to 3.6 according to [NEP 29](https://numpy.org/neps/nep-0029-deprecation_policy.html). We should bump to 3.7, but many libraries (torch!) are relatively slow to deprecate python versions, this seems like a reasonable compromise which allows us to type the library.

## Questions
 - [ ] Considering the range of inputs for contract/contract path, I'm leaning towards putting an Any on them for now. Any objections?
 - [ ] ArrayType is a Union which could *become* a `np.ndarray` meaning all interfaces should use `np.asanarray` upon call to convert to `np.ndarray` types. It doesn't seem `asanyarray` is typed yet in the docs. Considering leaving these as Any as well since we tend to pass the base type through (called Tensor classes). Any guidance here?

## Todos
  - [x] Bump Python to >= 3.6
  - [x] Add Python typing to the library
  - [x] Add a MyPy merge gate

## Status
- [ ] Ready to go

### DIFF
--- a/.github/workflows/Linting.yml
+++ b/.github/workflows/Linting.yml
@@ -1,4 +1,4 @@
-name: Tests
+name: Linting
 
 on:
   push:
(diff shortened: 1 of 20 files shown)


jcmgray (approved):
All seems OK to me! I don't have much experience with python's typing however to offer any in-depth review.

dgasmith:
Thanks for the review @jcmgray. I think this is moving in the right direction, but we will need to revisit the typing interface for some of the more complex areas (dynamic programming, the diverse input types into contract, etc) before we put out a release.

PR #174, opened 2021-11-06
Prevents double contraction of floats in dynamic path, fixes #172

### LINKED ISSUE #172: Bug: Dynamic_programming path solver may return illegal contraction path
Tested on GitHub master HEAD.

```python
oem.paths.dynamic_programming([{0,1}, {}], {0, 1}, {0:2, 1:2})
[(1,), (0, 1)] # sth. wrong here
oem.paths.greedy([{0,1}, {}], {0, 1}, {0:2, 1:2})
[(0, 1)] # other solvers return correct result
```

Though such tensor contraction case is weird and a very corner case, it indeed can occur in https://github.com/google/TensorNetwork/blob/master/tensornetwork/quantum/quantum.py

dgasmith:
## Description
The dynamic path will find disconnected subgraphs such as `aa, bc->bc` and automatically contract `aa->` on the fly as a pre optimization measure; however, the contraction `,bc->bc` (float, matrix -> matrix) was not considered and the float is first pre contracted `->` (a no-op). This PR prevents these no-ops to simplify the output path.

## Status
- [x] Ready to go

### DIFF
--- a/opt_einsum/paths.py
+++ b/opt_einsum/paths.py
@@ -1016,7 +1016,7 @@ class DynamicProgramming(PathOptimizer):
                 cost_cap = self.cost_cap
             # set the factor to increase the cost by each iteration (ensure > 1)
             if len(subgraph_inds) == 0:
-                cost_increment = 1
+                cost_increment = 2
             else:
                 cost_increment = max(min(map(size_dict.__getitem__, subgraph_inds)), 2)
 
--- a/opt_einsum/paths.py
+++ b/opt_einsum/paths.py
@@ -861,7 +861,7 @@ def _dp_parse_out_single_term_ops(inputs, all_inds, ind_counts):
     inputs_parsed, inputs_done, inputs_contractions = [], [], []
     for j, i in enumerate(inputs):
         i_reduced = i - i_single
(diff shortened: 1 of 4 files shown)


opt_einsum/paths.py
```diff
@@ -1015,7 +1015,10 @@ def __call__(self,
             else:
                 cost_cap = self.cost_cap
             # set the factor to increase the cost by each iteration (ensure > 1)
-            cost_increment = max(min(map(size_dict.__getitem__, subgraph_inds)), 2)
+            if len(subgraph_inds) == 0:
+                cost_increment = 1
```
dgasmith, line 1019: @jcmgray is this value correct?

jcmgray: I think shouldn't matter as subgraph will always be size 1? But probably should be `2`, (since its a multiplicative increment 1x does nothing).

jcmgray (approved):

opt_einsum/paths.py
dgasmith, line 1019:
```suggestion
                cost_increment = 2
```

PR #176, opened 2021-11-22
"Finalize" MyPy

dgasmith:
## Description
For the upcoming release, I did wish to attempt to finalize a round of MyPy, that is, all functions typed at least in a generic way. There still exist too many `Any` types, but we're making headway. In particular, typing the paths proved to be instructive and helped to unpack some type thrashing. This also helped improve the readability of the dynamic path slightly; I still wish to do another pass or two to make it more self-evident what is happening as this codebase matures.

The last item I wish to do in this PR is switch from YAPF to the black formatter. Black is raising in popularity over the past few years and generally find it much more stable than YAPF. A test run only shows ~50 lines of reasonable changed beyond mapping `' -> "`, but helps tremendously for our length arg-based function calls.

As a note, I'm not adding a `py.typed` yet- I would like to handle the ArrayType better than we currently do before I declare the types "ready".

## Questions
- [x] @jcmgray any objections to using the Black formatter over YAPF?

## Status
- [x] Ready to go


jcmgray (approved):
Looks good to me. I haven't generally used python typing hints much so far, but would 'protocols' be the best way to do e.g. an ArrayLike? just read this: https://www.daan.fyi/writings/python-protocols from which it seems something like this might be good:

```python
from typing import Protocol, Tuple

class ArrayLike(Protocol):
    shape: Tuple[int]  # or maybe some less strict
```
(EDIT: I notice this is only available for python 3.8+)

Regarding yapf to black, seems like good choice to me! especially given its wide uptake now.

dgasmith:
There are NumPy types that we should use: https://github.com/numpy/numpy/blob/main/numpy/typing/_array_like.py

Our NumPy minimum version is quite old, let me see if we can do anything to improve the situation in another PR.

Issue #171, opened 2021-10-23
Customized cost functions for faster contractions


jcmgray:
As part of some work on contracting some very large tensor networks we found that a good heuristic cost to optimize for is not just 'FLOPs' but also 'WRITE' - i.e. the size of the intermediate output tensors regardless of the cost to product it. 

For example, an outer product with shapes like `(d, 1), (1, d)->(d, d)` has the same WRITE cost as matrix multiplication with shapes like `(d, d), (d, d) -> (d, d)` despite having a FLOPs cost `d` times lower. 

If you take a cost as `FLOPS + alpha * WRITE` and optimize for this you can get some significant and fairly generic gains across hardware in terms of actual efficiency and contraction time. Relevant probably for e.g. #103.

Here (you can ignore what the contraction is and the 'index sorting' bit), we have FLOPS and WRITE on top and actual contraction time on bottom on a single cpu thread:

![cpu-alpha-efficiency](https://user-images.githubusercontent.com/8982598/138567608-e23b6da2-cddf-4bc6-a54e-abb5c09d022d.png)

You can see there's about a 5x speedup.

Here's the same for a larger contraction on an A100 (big fast gpu) with a speedup of 8x:

![gpu-alpha-efficiency](https://user-images.githubusercontent.com/8982598/138567727-532512f8-77d5-484b-bc15-e73d7408e84e.png)

In both cases choosing `alpha=64` is very decent despite probably having different ratios of cpu cycles to memory speed etc etc. Generally there seems to be a window where neither the FLOPS or WRITE cost is compromised too much. **EDIT: I should emphasize that the above contractions tend to produce very low arithmetic intensity contractions, so the speedups will not always be this large.**

I've had this in my own branch of `opt_einsum` for a while (just for the `dp` optimizer, but which one can use to take *any* large contraction path/tree and iteratively 'reconfigure' subtrees of), but it would be nice to have it in the main `opt_einsum` and it also might be useful to others.

Currently the interface is like:
```python
# use the default alpha (probably 64)
opt = DynamicProgramming(minimize='combo')

# use a specific alpha
opt = DynamicProgramming(minimize='combo-100')
```
@dgasmith, thoughts on including?

jcmgray:
These are now available with #181.

PR #181, opened 2022-01-15
more custom dp scores for faster contractions

jcmgray:
## Description

This adds more custom functions to the dp solver as outlined in #171. For example:
```python
opt = oe.DynamicProgramming(minimize='combo')
```
should give better real world performance in many cases than the pure minimized flop cost alone.
Also included is a couple of 'micro' optimizations  for the path finders performance - namely, making some inner loop calls functions not methods.

## Todos
  - [ ] at some point might be nice to add these to the random greedy path as well

## Questions
- [x]  ~~I tried the black formatter and it showed lots of changes across many files - currently have just committed those on the files modified...? Am I invoking it wrong? (edit: might just be that the conda `black` is a bit out-of-date)~~

## Status
- [x] Ready to go

### DIFF
--- a/opt_einsum/paths.py
+++ b/opt_einsum/paths.py
@@ -60,10 +60,7 @@ class PathOptimizer:
     """
 
     def _check_args_against_first_call(
-        self,
-        inputs: List[ArrayIndexType],
-        output: ArrayIndexType,
-        size_dict: Dict[str, int],
+        self, inputs: List[ArrayIndexType], output: ArrayIndexType, size_dict: Dict[str, int],
     ) -> None:
         """Utility that stateful optimizers can use to ensure they are not
         called with different contractions across separate runs.
@@ -169,10 +166,7 @@ def calc_k12_flops(
 
 
 def _compute_oversize_flops(
-    inputs: Tuple[FrozenSet[str]],
-    remaining: List[ArrayIndexType],
-    output: ArrayIndexType,
-    size_dict: Dict[str, int],
+    inputs: Tuple[FrozenSet[str]], remaining: List[ArrayIndexType], output: ArrayIndexType, size_dict: Dict[str, int],
 ) -> int:
     """
     Compute the flop count for a contraction of all remaining arguments. This
@@ -185,10 +179,7 @@ def _compute_oversize_flops(
 
 
 def optimal(
-    inputs: List[ArrayIndexType],
-    output: ArrayIndexType,
-    size_dict: Dict[str, int],
-    memory_limit: Optional[int] = None,
+    inputs: List[ArrayIndexType], output: ArrayIndexType, size_dict: Dict[str, int], memory_limit: Optional[int] = None,
 ) -> PathType:
     """
     Computes all possible pair contractions in a depth-first recursive manner,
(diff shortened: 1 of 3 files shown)


opt_einsum/paths.py
```diff
@@ -971,14 +970,112 @@ def _dp_compare_size(
...
-    mem = helpers.compute_size_by_dict(i, size_dict)
+    mem = compute_size_by_dict(i, size_dict)
     cost = max(cost1, cost2, mem)
     if cost <= cost_cap:
         if s not in xn or cost < xn[s][1]:
             if memory_limit is None or mem <= memory_limit:
                 xn[s] = (i, cost, (contract1, contract2))
 
 
+def _dp_compare_write(
```
dgasmith, line 981: These functions have surpassed what is reasonable to expect for straight args. Can we either pass as kwargs, a named tuple, or a data class?

Please type them as well.

opt_einsum/paths.py
```diff
@@ -971,14 +970,112 @@ def _dp_compare_size(
...
+
+minimize_finder = re.compile("(flops|size|write|combo|limit)-*(\d*)")
+
+
+@functools.lru_cache(128)
+def _parse_minimize(minimize):
+    """This works out what local scoring function to use for the dp algorithm
+    as well as a `naive_scale` to account for the memory_limit checks.
+    """
+    if minimize == "flops":
```
dgasmith, line 1056: Please change to `if/elif/else` patterns here. Functionality it doesn't matter, but helps with readability.

opt_einsum/tests/test_paths.py
```diff
@@ -270,6 +270,26 @@ def test_custom_dp_can_set_cost_cap():
...
+        ("combo-256", 983790, 2016),
+        ("limit-256", 983832, 2016),
+    ],
+)
+def test_custom_dp_can_set_minimize(minimize, cost, width):
+    eq, shapes = oe.helpers.rand_equation(10, 4, seed=43)
+    opt = oe.DynamicProgramming(minimize=minimize)
+    info = oe.contract_path(eq, *shapes, shapes=True, optimize=opt)[1]
+    assert info.opt_cost == cost
+    assert info.largest_intermediate == width
```
dgasmith, line 290: Can we double check the path itself as well?

dgasmith (commented):
Looks great! Can we add a few notes on the DP optimizer's new minimizers in the docs?

opt_einsum/paths.py
jcmgray: In general I'd agree, but I do think in this one case, i.e. a tight loop where performance is very important (this exact place is the bottleneck for many codes), using plain args is the right call. E.g. even just calling the function with kwargs slows down the dp optimizer a few percent - more 'readable' options like named tuples would incur a higher overhead. In fact given all the variable names already match the function signature I'm personally not sure it would be clearer either?

I had missed that all these functions had been type hinted when rebasing though - will add those.

dgasmith (approved):
LGTM once mypy is worked through. Let me know if you would like a hand!

opt_einsum/paths.py
```diff
@@ -971,14 +970,115 @@ def _dp_compare_size(
...
+        # as otherwise memory_limit check can cause problems
+        return minimize, float("inf")
+
+    # parse out a customized value for the combination factor
+    match = minimize_finder.fullmatch(minimize)
+    if match is None:
+        raise ValueError(f"Couldn't parse `minimize` value: {minimize}.")
+
+    minimize, factor = match.groups()
+    factor = float(factor) if factor else DEFAULT_COMBO_FACTOR
```
jcmgray, line 1073: @dgasmith this is the line that mypy was complaining about:
```
Incompatible types in assignment (expression has type "float", variable has type "str")
```
 If you recognise the problem let me know! or I will experiment fixing.

jcmgray: OK think its just a matter of not reusing the variable name..

PR #195, opened 2022-07-05
Fix `dp` to work for contractions with no sized axes (scalars only).

romanngg:
Fix `dp` to work for contractions with no sized axes (scalars only).


jcmgray (approved):
Thanks, all looks good to me!

dgasmith (approved):
LGTM, ty.

PR #196, opened 2022-07-06
allow backend=None

### LINKED ISSUE #193: allow `backend=None` as shorthand for 'auto'
A couple of times I've be caught out (e.g. https://github.com/jcmgray/quimb/issues/130) that the special value for inferring the backend is not `backend=None`. Might be nice to allow both?

If so would just require a simple:
```python
if backend not in ("auto", None):
    return backend
```
in `parse_backend` of `contract.py`.

jcmgray:
## Description
Allows `backend=None`, which should be fairly unambiguous, as an alternative to 'auto', closes #193.

## Status
- [x] Ready to go

### DIFF
--- a/opt_einsum/contract.py
+++ b/opt_einsum/contract.py
@@ -550,7 +550,7 @@ def parse_backend(arrays: Sequence[ArrayType], backend: str) -> str:
     """Find out what backend we should use, dipatching based on the first
     array if ``backend='auto'`` is specified.
     """
-    if backend != "auto":
+    if backend not in ("auto", None):
         return backend
     backend = infer_backend(arrays[0])
 
--- a/opt_einsum/contract.py
+++ b/opt_einsum/contract.py
@@ -4,6 +4,7 @@ Contains the primary optimization and contraction routines.
 
 from collections import namedtuple
 from decimal import Decimal
+from functools import lru_cache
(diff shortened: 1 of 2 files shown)


opt_einsum/contract.py
```diff
@@ -550,7 +550,7 @@ def parse_backend(arrays: Sequence[ArrayType], backend: str) -> str:
     """Find out what backend we should use, dipatching based on the first
     array if ``backend='auto'`` is specified.
     """
-    if backend != "auto":
+    if backend not in ("auto", None):
```
dgasmith, line 553: Can you change the types? `backend: Optional[str]` on line 549 and likely elsewhere.

jcmgray: Yes - have done now

jcmgray:
Have updated types, though I see the `mypy` check failed - will try and work out what's going on there.

I've also included a little optimization of `infer_backend` (1us -> 500ns) making use of `lru_cache` - let me know if for some reason this doesn't work! (I think now we don't support older pythons should be fine).

dgasmith (approved):
We follow NumPy's python support recommendations. We're up to 3.7 minimum so it shouldn't be a problem.

This LGTM!

PR #197, opened 2022-07-17
Support for Python array API standard

IsaacBreen:
## Description
This PR adds support for libraries that conform to the [Python array API standard](https://data-apis.org/array-api/latest/index.html). It discovers complying modules upon import of opt_einsum.

Eventually (and hopefully soon), once the array API standard becomes ubiquitous, this should largely eliminate the need for backend-specific code.

## Todos
  - [x] Tests
  - [ ] Update docs

## Status
- [x] Ready to go

### DIFF
--- a/opt_einsum/backends/array_api.py
+++ b/opt_einsum/backends/array_api.py
@@ -11,13 +11,14 @@ import numpy as np
 from ..sharing import to_backend_cache_wrap
 
 
-def discover_array_api_eps():
-    """Discover array API backends and return their entry points."""
+def discover_array_apis():
+    """Discover array API backends."""
     if sys.version_info >= (3, 10):
-        return entry_points(group='array_api')
+        eps = entry_points(group='array_api')
     else:
         # Deprecated - will raise warning in Python versions >= 3.10
-        return entry_points().get('array_api', [])
+        eps = entry_points().get('array_api', [])
+    return [ep.load() for ep in eps]
 
 def make_to_array(array_api: ModuleType) -> Callable:
     """Make a ``to_[array_api]`` function for the given array API."""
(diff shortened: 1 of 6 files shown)


dgasmith (commented):
This is going to take a bit to go through and I'll be out next week. Looks a truly fantastic PR!

dgasmith:
Looks like plenty of errors for Py3.6 where `importlib.metadata` is not yet extant.

IsaacBreen:
Thanks Daniel! The errors for Python <= 3.7 should be fixed. I've set it to bypass array API backend discovery in such cases. Numpy doesn't support <=3.7 as-is, so I don't think this will be missed.

I've fixed up the formatting too.

opt_einsum/backends/array_api.py
```diff
@@ -0,0 +1,72 @@
...
+
+
+def make_build_expression_function(array_api: ModuleType) -> Callable:
+    """Make a ``build_expression`` function for the given array API."""
+
+    def build_expression(_, expr):  # pragma: no cover
+        """Build an array API function based on ``arrays`` and ``expr``."""
+
+        def array_api_contract(*arrays):
+            return expr._contract([make_to_array_function(array_api)(x) for x in arrays], backend=array_api.__name__)
```
jcmgray, line 48: Should `to_array = make_to_array_function(array_api)` maybe be assigned in the outer scope to avoid reconstructing it for every array? (and below)

opt_einsum/backends/array_api.py
```diff
@@ -40,23 +40,26 @@ def to_array(array):  # pragma: no cover
...
 def make_build_expression_function(array_api: ModuleType) -> Callable:
     """Make a ``build_expression`` function for the given array API."""
+    _to_array_api = to_array_api[array_api.__name__]
 
     def build_expression(_, expr):  # pragma: no cover
         """Build an array API function based on ``arrays`` and ``expr``."""
 
         def array_api_contract(*arrays):
-            return expr._contract([make_to_array_function(array_api)(x) for x in arrays], backend=array_api.__name__)
+            return expr._contract([_to_array_api(x) for x in arrays], backend=array_api.__name__)
```
IsaacBreen, line 49: # # @jcmgray How about this?

opt_einsum/backends/array_api.py
IsaacBreen: Yes, that's much better

opt_einsum/backends/array_api.py
```diff
@@ -0,0 +1,75 @@
...
+import sys
+from typing import Callable
+from types import ModuleType
+
+import numpy as np
+
+from ..sharing import to_backend_cache_wrap
+
+
+def discover_array_apis():
```
dgasmith, line 13: Possible to type this function?

opt_einsum/backends/dispatch.py
```diff
@@ -122,6 +123,7 @@ def has_tensordot(backend: str) -> bool:
     "cupy": _cupy.build_expression,
     "torch": _torch.build_expression,
     "jax": _jax.build_expression,
+    **_array_api.build_expression,
```
dgasmith, line 126: Is there any way to check if `Jax` adds an array interface so that we can catch it and test before hand?

opt_einsum/backends/array_api.py
```diff
@@ -0,0 +1,75 @@
...
+
+        if sys.version_info >= (3, 10):
+            eps = entry_points(group="array_api")
+        else:
+            # Deprecated - will raise warning in Python versions >= 3.10
+            eps = entry_points().get("array_api", [])
+        return [ep.load() for ep in eps]
+    else:
+        # importlib.metadata was introduced in Python 3.8, so it isn't available here. Unable to discover any array APIs.
+        return []
```
dgasmith, line 26: Numpy does not officially support below 3.8: https://numpy.org/neps/nep-0029-deprecation_policy.html

It would be worth considering dropping Python 3.7 as well. @jcmgray what do you think?

dgasmith (commented):
Apologies for the delay here, this appears quite polished. From your view, what needs to happen to get this in?

dgasmith:
@IsaacBreen I'm looking to push out another release in the next month or so. Is this a PR which you would like to support?

dgasmith:
@NeilGirdhar I have not, but I'm happy to take a PR!

PR #200, opened 2022-09-26
[BUG FIX] Add shapes support to parse_einsum_input

janeyx99:
## Description
This PR adds support for the shapes kwarg for `parse_einsum_input`. This also fixes a bug where `contract_path` takes in the shapes parameter but `parse_einsum_input` is called without respecting the shapes parameter.

The bug results because even though the user had input shapes as True, `contract_path` will call `parse_einsum_input` _before_ caring about shapes, and within `parse_einsum_input`, will call .shape on an array that is already a shape.

## Todos
  - [x] Add support for the shapes kwarg in parse_einsum_input
  - [x] Pass along the shapes kwarg in the call from contract_path

## Questions
- [x] The lint failure looks to be infra flakiness--how can I rerun?

## Status
- [x] Ready to go

### DIFF
--- a/opt_einsum/parser.py
+++ b/opt_einsum/parser.py
@@ -3,7 +3,7 @@ A functionally equivalent parser of the numpy.einsum input parser
 """
 
 import itertools
-from typing import Any, Dict, Iterator, List, Tuple
+from typing import Any, Dict, Iterator, List, Tuple, Union
 
 import numpy as np
 
@@ -224,7 +224,7 @@ def convert_subscripts(old_sub: List[Any], symbol_map: Dict[Any, Any]) -> str:
     return new_sub
 
 
-def convert_interleaved_input(operands: List[Any]) -> Tuple[str, List[Any]]:
+def convert_interleaved_input(operands: Union[List[Any], Tuple[Any]]) -> Tuple[str, List[Any]]:
     """Convert 'interleaved' input to standard einsum input."""
     tmp_operands = list(operands)
     operand_list = []
@@ -262,7 +262,7 @@ def convert_interleaved_input(operands: List[Any]) -> Tuple[str, List[Any]]:
     return subscripts, operands
 
 
-def parse_einsum_input(*operands: Any, shapes: bool = False) -> Tuple[str, str, List[ArrayType]]:
+def parse_einsum_input(operands: Any, shapes: bool = False) -> Tuple[str, str, List[ArrayType]]:
     """
     A reproduction of einsum c side einsum parsing in python.
 
--- a/opt_einsum/parser.py
+++ b/opt_einsum/parser.py
@@ -262,9 +262,15 @@ def convert_interleaved_input(operands: List[Any]) -> Tuple[str, List[Any]]:
     return subscripts, operands
 
 
-def parse_einsum_input(operands: Any) -> Tuple[str, str, List[ArrayType]]:
+def parse_einsum_input(operands: Any, **kwargs: Any) -> Tuple[str, str, List[ArrayType]]:
     """
     A reproduction of einsum c side einsum parsing in python.
+
+    **Parameters:**
+    Intakes the same inputs as `contract_path`, but NOT the keyword args. The only
+    supported keyword argument is:
+    - **shapes** - *(bool, optional)* Whether ``parse_einsum_input`` should assume arrays (the default) or
+        array shapes have been supplied.
 
     Returns
     -------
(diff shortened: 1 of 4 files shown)


opt_einsum/parser.py
```diff
@@ -262,10 +262,16 @@ def convert_interleaved_input(operands: List[Any]) -> Tuple[str, List[Any]]:
     return subscripts, operands
 
 
-def parse_einsum_input(operands: Any) -> Tuple[str, str, List[ArrayType]]:
+def parse_einsum_input(operands: Any, **kwargs: Any) -> Tuple[str, str, List[ArrayType]]:
```
dgasmith, line 265: Is this is in an internal function which does not have pass through kwargs, we can use `shapes: bool = False` rather than the kwarg unpacking to skip lines `297-302`.

dgasmith (commented):
Looks good to me- can we please add a test to validate the functionality?

Asking @jcmgray to review as the original implementer here.

opt_einsum/parser.py
```diff
@@ -262,10 +262,16 @@ def convert_interleaved_input(operands: List[Any]) -> Tuple[str, List[Any]]:
     return subscripts, operands
 
 
-def parse_einsum_input(operands: Any) -> Tuple[str, str, List[ArrayType]]:
+def parse_einsum_input(*operands: Any, shapes: bool = False) -> Tuple[str, str, List[ArrayType]]:
```
dgasmith, line 265:
```suggestion
def parse_einsum_input(operands: Any, shapes: bool = False) -> Tuple[str, str, List[ArrayType]]:
```

We shouldn't have to change the operands unpacking here. The logs are pretty long, but I suspect this is the key issue.

dgasmith:
> Welp, lol, messing with the parameters did not go well--I'll take another look tomorrow but let me know if you have any advice--I kept getting errors when I didn't use *operands about shapes getting multiple inputs :c.

That is very odd- can you post the exact error message?

opt_einsum/parser.py
janeyx99: Yea, I think my latest change should fix this

janeyx99:
> That is very odd- can you post the exact error message?

We are good now!

opt_einsum/tests/test_parser.py
```diff
@@ -12,3 +13,21 @@ def test_get_symbol():
     assert get_symbol(55295) == "\ud88b"
     assert get_symbol(55296) == "\ue000"
     assert get_symbol(57343) == "\ue7ff"
+
+
+def test_parse_einsum_input():
+    eq = "ab,bc,cd"
+    ops = [np.random.rand(2, 3), np.random.rand(3, 4), np.random.rand(4, 5)]
+    input_subscripts, output_subscript, operands = parse_einsum_input([eq, *ops], shapes=True)
```
dgasmith, line 21: My understanding is that this test shouldn't work as we expect tuples of ints rather than tensors. I must be missing something here however.

dgasmith (commented):
Very close! One last question.

opt_einsum/tests/test_parser.py
janeyx99: Oh that is a very good point.

This also just works because parse_einsum_input doesn't do anything to the operands (whether they be shapes or operands) and should just pass them along.

janeyx99: I've modified the test so it tests the intended behavior.

jcmgray (approved):
All looks good to me, thanks for the PR @janeyx99 and adding tests etc.!

opt_einsum/tests/test_parser.py
dgasmith: We have a faulty test in this case, It just so happens that `len(arr)` provides the shape of the first dimension. We need to ensure that a corresponding test fails for the given test case. A fairly minor point, but the tests as written can be misleading in the future.

janeyx99:
Is there anything on my end I should do to merge this PR?

dgasmith:
I'll recap from a comment above:
> We have a faulty test in this case, It just so happens that len(arr) provides the shape of the first dimension. We need to ensure that a corresponding test fails for the given test case. A fairly minor point, but the tests as written can be misleading in the future.

I think we need to make sure the test properly raises if an array is passed as a shape. Either that or detect automatically if the object has a `.shape` attribute or not.

janeyx99:
Ah oops, I didn't see that last comment. I've added a test case now.

janeyx99:
Lint failure looks unrelated

dgasmith:
Looks like MyPy bumped, would you mind fixing? Removing the `"type: ignore"` from line `opt_einsum/paths.py:1251` should patch it.

opt_einsum/tests/test_parser.py
```diff
@@ -24,6 +24,19 @@ def test_parse_einsum_input():
     assert operands == ops
 
 
+def test_parse_einsum_input_shapes_error():
+    eq = "ab,bc,cd"
+    ops = [np.random.rand(2, 3), np.random.rand(3, 4), np.random.rand(4, 5)]
+
+    try:
```
dgasmith, line 31: For pytest you can use `with pytest.raises(ValueError)` here: https://docs.pytest.org/en/7.1.x/how-to/assert.html

dgasmith (commented):
Very close! This is a deceptively tricky PR.

opt_einsum/tests/test_parser.py
janeyx99: Ah, I just avoided pulling in dependencies.

dgasmith:
Amazing, thank you!

Issue #202, opened 2022-10-28
When I use opt_einsum optimizes torch.einum, the running time after optimization increases.


edwin-zft:
```
import numpy as np
import time
import torch
from opt_einsum import contract

dim = 4

x = torch.randn(6 ,4 ,4, 4)
w1 = torch.randn(1,4,dim)
w2 = torch.randn(dim,4,dim)
w3 = torch.randn(dim,4,dim)
w4 = torch.randn(dim,8,dim)
w5 = torch.randn(dim,8,dim)
w6 = torch.randn(dim,4,1)

def naive(x, w1, w2, w3, w4, w5, w6):
    return torch.einsum('bkxy,ikj,jxm,myf,fpl,lqz,zri -> bpqr', x, w1, w2, w3, w4, w5, w6)

def optimized(x, w1, w2, w3, w4, w5, w6):
    return contract('bkxy,ikj,jxm,myf,fpl,lqz,zri -> bpqr', x, w1, w2, w3, w4, w5, w6)

```
**The respective running time:**
```
naive
0.0005145072937011719
```
```
optimized
0.876018762588501
```
I want to know what caused this.Thanks!

jcmgray:
Hey @edwin-zft, I get:
```python
%%timeit
y = naive(x, w1, w2, w3, w4, w5, w6)
# 536 µs ± 4.06 µs per loop (mean ± std. dev. of 7 runs, 1,000 loops each)
```
vs.
```python
%%timeit
y = optimized(x, w1, w2, w3, w4, w5, w6)
# 470 µs ± 2.07 µs per loop (mean ± std. dev. of 7 runs, 1,000 loops each)
```
and as a bonus:
```python
expr = contract_expression(
    'bkxy,ikj,jxm,myf,fpl,lqz,zri->bpqr', 
    x.shape, w1.shape, w2.shape, w3.shape, w4.shape, w5.shape, w6.shape,
    optimize='dp',
)

%%timeit
y = expr(x, w1, w2, w3, w4, w5, w6)
# 72.2 µs ± 758 ns per loop (mean ± std. dev. of 7 runs, 10,000 loops each)
```
so maybe its just a warm-up issue for you, are you using `timeit`?

edwin-zft:
Thank you for your reply！
I reuse `timeit `for testing
```
%%timeit
y = naive(x, w1, w2, w3, w4, w5, w6)
#0.00007126200944185257s (10000 loops)
```
vs.
```
%%timeit
[...]

PR #205, opened 2022-11-07
Disable torch.backends.opt_einsum to avoid duplicate work

janeyx99:
## Description
Recently, `torch.einsum` has improved to automatically optimize for multi-contractions if the opt_einsum library is installed. This way, torch users can reap benefits easily. However, this change may inadvertently cause regressions for those who use `opt_einsum.contract` with torch tensors. While these use cases can be improved by just changing `opt_einsum.contract` calls to `torch.einsum`, it is not right to cause silent regressions. 

Consider the following example:
```
import torch
import opt_einsum

A = torch.rand(4, 5)
B = torch.rand(5, 6)
C = torch.rand(6, 7) 
D = opt_einsum.contract('ij,jk,kl->il', A, B, C)
```

Looking through the code, `opt_einsum.contract` will figure out an ideal path + return that path in the form of contraction tuples (commonly pairs). Then, for each of the contraction tuples, `opt_einsum` may call BACK into the torch backend and call `torch.einsum`.

Since the contractions are commonly pairs, calling back into torch.einsum will do what it used to--just directly do the contraction. 

HOWEVER. There are cases where the contractions are not necessarily pairs (@dgasmith had mentioned that hadamard products are faster when chained vs staggered in pairs) and in this case, `torch.einsum` will do unnecessary work in recomputing an ideal path, which is a regression from the previous state.

This PR simply asks "hey, does the opt_einsum backend exist in torch?" If so, this means `torch.einsum` will do the unnecessary work in recomputing an ideal path, so let's just turn it off.

## Todos
  - [x] Add the code to disable torch's opt_einsum optimization
  - [ ] Test cases? 

## Questions
- [ ] Unsure how to best test this/what use case would ensure that, without this code, the path would be computed more than once, and after this code, the path would only be computed at the top level. We could just test that torch.backends.opt_einsum.enabled is globally False, I guess.

## Status
- [ ] Ready to go

### DIFF
--- a/opt_einsum/backends/torch.py
+++ b/opt_einsum/backends/torch.py
@@ -6,6 +6,7 @@ import numpy as np
 
 from ..parser import convert_to_valid_einsum_chars
 from ..sharing import to_backend_cache_wrap
+from functools import lru_cache
 
 __all__ = [
     "transpose",
@@ -40,18 +41,27 @@ def transpose(a, axes):
     """Normal torch transpose is only valid for 2D matrices."""
     return a.permute(*axes)
 
+@lru_cache(None)
+def get_einsum():
+    torch, _ = _get_torch_and_device()
+    if hasattr(torch, "_VF") and hasattr(torch._VF, "einsum"):
+        print("called")
+        return torch._VF.einsum
+
+    def einsum_no_opt(*args, **kwargs):
+        if hasattr(torch, "backends") and hasattr(torch.backends, "opt_einsum"):
+            with torch.backends.opt_einsum.flags(enabled=False):
+                return torch.einsum(*args, **kwargs)
+        return torch.einsum(*args, **kwargs)
+    return einsum_no_opt
+
 
 def einsum(equation, *operands):
     """Variadic version of torch.einsum to match numpy api."""
     # rename symbols to support PyTorch 0.4.1 and earlier,
     # which allow only symbols a-z.
     equation = convert_to_valid_einsum_chars(equation)
-
-    torch, _ = _get_torch_and_device()
-    # disable opt_einsum in torch so that it does not recompute any path
-    if hasattr(torch, "backends") and hasattr(torch.backends, "opt_einsum"):
-        torch.backends.opt_einsum.enabled = False
-    return torch.einsum(equation, operands)
+    return get_einsum()(equation, operands)
 
 
 def tensordot(x, y, axes=2):
--- a/opt_einsum/backends/torch.py
+++ b/opt_einsum/backends/torch.py
@@ -46,16 +46,18 @@ def transpose(a, axes):
 def get_einsum():
     torch, _ = _get_torch_and_device()
     if hasattr(torch, "_VF") and hasattr(torch._VF, "einsum"):
-        print("called")
+        print("returning VF")
         return torch._VF.einsum
-    print("did not get called")
 
+    if not hasattr(torch, "backends") or not hasattr(torch.backends, "opt_einsum"):
+        print("returning normal torch")
+        return torch.einsum
+    
     def einsum_no_opt(*args, **kwargs):
-        if hasattr(torch, "backends") and hasattr(torch.backends, "opt_einsum"):
-            with torch.backends.opt_einsum.flags(enabled=False):
-                return torch.einsum(*args, **kwargs)
-        return torch.einsum(*args, **kwargs)
-
+        with torch.backends.opt_einsum.flags(enabled=False):
+            return torch.einsum(*args, **kwargs)
+    
+    print("returning normal torch with opt off")
     return einsum_no_opt
 
 
--- a/opt_einsum/backends/torch.py
+++ b/opt_einsum/backends/torch.py
@@ -48,6 +48,9 @@ def einsum(equation, *operands):
     equation = convert_to_valid_einsum_chars(equation)
 
     torch, _ = _get_torch_and_device()
+    # disable opt_einsum in torch so that it does not recompute any path
+    if hasattr(torch, "backends") and hasattr(torch.backends, "opt_einsum"):
+        torch.backends.opt_einsum.enabled = False
     return torch.einsum(equation, operands)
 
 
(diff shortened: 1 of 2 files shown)


janeyx99:
tagging @dgasmith to get it on the radar

dgasmith:
@janeyx99 Apologies, I'm traveling for the next week and will be slow to respond.

@jcmgray Can you get eyes on this?

jcmgray:
Hi! I'm guessing `torch` does not have a way to turn off the path optimization at call time (e.g. `torch.einsum(..., optimize=False)`)? That would obviously be ideal.

My thoughts would be the following:

- it's rare to come across paths where not contracting pairwise is *practically* beneficial (theoretically ops-wise pairwise is always optimal). I think currently only imposing a size limit makes `opt_einsum` produce these?
- In the case that say a 3 tensor contraction is encountered, the path optimization would only run on the those 3, not the whole expression - i.e. unlikely to have much overhead

So in general my opinion is that is unlikely to make a difference for most cases, or even be beneficial. Other considerations:

- for testing and debugging, it is useful to be able to turn off the path (see e.g. #202)
- some users might be calling `oe.contract` *and* `torch.einsum` and expect both to be fast, in which case it might be confusing that calling `opt_einsum` makes this global change that slows down `torch`.

jcmgray:
A bit hacky, but could we try and call `_VF.einsum` directly and skip the `torch` side python processing?

dgasmith:
@janeyx99 Thoughts here?

janeyx99:
My view on this is that maybe it’s not a big deal and so the complexity/additional check may not be worth it. That said, I would be happy to try calling _VF.einsum directly if that is preferable to the context manager.

jcmgray:
I think it's a rare and small enough effect that I'd be happy to ignore. Having said that:

1) I suppose to respect `size_limit`, calling the raw `einsum` might be important.
2) The time difference is small but not zero:

```python
import torch
from torch import _VF

x = torch.rand((2,))
y = torch.rand((2,))
z = torch.rand((2,))
eq = "a,a,a->a"

%timeit torch.einsum(eq, x, y, z)
# 37.3 µs ± 70.1 ns per loop (mean ± std. dev. of 7 runs, 10,000 loops each)

%%timeit
with torch.backends.opt_einsum.flags(enabled=False):
    torch.einsum(eq, x, y, z)
# 8.94 µs ± 95.4 ns per loop (mean ± std. dev. of 7 runs, 100,000 loops each)

%timeit _VF.einsum(eq, (x, y, z))
# 3.86 µs ± 36 ns per loop (mean ± std. dev. of 7 runs, 100,000 loops each)
```

we could do something like:

```python
from functools import lru_cache

@lru_cache(None)
def get_einsum():
    try:
        from torch import _VF
        return _VF.einsum
    except ImportError:  # maybe other errors, AttributeError?
        import torch

        def einsum_no_opt(*args, **kwargs):
            with torch.backends.opt_einsum.flags(enabled=False):
                return torch.einsum(*args, **kwargs)

        return einsum_no_opt
```

that way if relying on `_VF.einsum` breaks somehow, we have a sensible fallback.

janeyx99:
@dgasmith looks like there are conda package conflicts that aren't related to the PR

opt_einsum/backends/torch.py
```diff
@@ -41,14 +42,29 @@ def transpose(a, axes):
...
+@lru_cache(None)
+def get_einsum():
+    torch, _ = _get_torch_and_device()
+    if hasattr(torch, "_VF") and hasattr(torch._VF, "einsum"):
+        print("called")
+        return torch._VF.einsum
+    print("did not get called")
+
+    def einsum_no_opt(*args, **kwargs):
+        if hasattr(torch, "backends") and hasattr(torch.backends, "opt_einsum"):
```
jcmgray, line 54: Minor comment - could this check be moved outside the function defn? I.e. if we can't find `_VF.einsum` or the opt_einsum flags, just return `torch.einsum` from `get_einsum`? Then it won't be checked every einsum call.

janeyx99: done

opt_einsum/backends/torch.py
```diff
@@ -41,14 +42,31 @@ def transpose(a, axes):
     return a.permute(*axes)
 
 
+@lru_cache(None)
+def get_einsum():
+    torch, _ = _get_torch_and_device()
+    if hasattr(torch, "_VF") and hasattr(torch._VF, "einsum"):
+        print("returning VF")
```
dgasmith, line 49: If we remove the prints, I think we're good to go1

janeyx99: Are the current test cases sufficient to verify this doesn't break anything?

dgasmith:
@janeyx99 Given PyTorch has likely moved on since this PR was made and deprecated older versions, what is optimal to do here?

janeyx99:
I'm willing to clean this PR up if it is still advantageous from the opt_einsum perspective, but it has been sufficiently long enough where maybe this doesn't matter so much.

dgasmith:
My understanding was this PR would be optimal from a Torch perspective. If it's no longer needed or if Torch's minimum supported version no longer requires this patch, we can safely close.

dgasmith:
@janeyx99 Ah- I would recommend having `PyTorch` skip `contract_path` when `len(operands) <= 2` as there is nothing the code can do. While we do optimize for this fast path within `opt_einsum` there is still some parsing overhead. By this small optimization PyTorch can lower `torch.einsum`'s latency in general.

janeyx99:
@dgasmith we already do that! Though I’m not sure how that’s relevant to the code path I had in mind above. I think it’s also sufficient to document that torch has already upstreamed opt_einsum though

dgasmith:
@janeyx99 Got it, then we only deal with the edge case where it's optimal to perform a three-product Hadamard or similar. The performance hit is likely quite minimal. Would you mind closing this PR and adding a note to the torch documentation that you should use the torch implementations instead?

janeyx99:
okay! closing this PR, and planning to work on PyTorch issue 127109 instead.

Issue #189, opened 2022-06-05
Difficulty in finding the optimal conraction when there is a scalar


Geositta2000:
The question in this issue is, suppose I have a contraction involving a scalar, `\sum_b 0.5 A[a,b] B[b,c] = C[a,c]`, in principle, the timing will be similar to  `\sum_b A[a,b] B[b,c] = C[a,c]`. I can find the similar timing, but only in auto/optimal searching path. All other algorithms don't work in the following example. In complicated contractions, the default searching path may not be sufficient. So I would like to ask if there is any approach can capture the good contraction when there is a scalar without going to the `optimal`. 

I may look for a walk around, namely, `np.multiply` the product array, since the cost will be O(N^2) than O(N^3). In the contraction when the intermediate dimension is much smaller then other dimensions, the wall time of `np.multiply` is comparable than the contraction timing. 

```
import numpy as np
import time
import opt_einsum as oe
                 

na = nc = 1000
nb = 5
n_iter = 100

A = np.random.random((na,nb))
B = np.random.random((nb,nc))


t_total = 0.
for i in range(n_iter):
    start = time.time()
    C = oe.contract('ij,jk->ik', A, B)
    end = time.time()
    t_total += end - start
print('AB->C',(t_total)/n_iter)



t_total = 0.
for i in range(n_iter):
    start = time.time()
    C = oe.contract(',ij,jk->ik', 0.5, A, B)
    end = time.time()
    t_total += end - start
print('AB->C scalar-auto',(t_total)/n_iter)

t_total = 0.
for i in range(n_iter):
    start = time.time()
    C = oe.contract(',ij,jk->ik', 0.5, A, B, optimize = 'dp')
    end = time.time()
    t_total += end - start
print('AB->C scalar-dp',(t_total)/n_iter)

t_total = 0.
for i in range(n_iter):
    start = time.time()
    C = oe.contract(',ij,jk->ik', 0.5, A, B, optimize = 'branch-all')
    end = time.time()
    t_total += end - start
print('AB->C scalar-branch-all',(t_total)/n_iter)

t_total = 0.
for i in range(n_iter):
    start = time.time()
    C = oe.contract(',ij,jk->ik', 0.5, A, B, optimize = 'branch-1')
    end = time.time()
    t_total += end - start
print('AB->C scalar-branch-1',(t_total)/n_iter)

t_total = 0.
for i in range(n_iter):
    start = time.time()
[...]

Issue #167, opened 2021-08-24
Is opt_einsum.paths.optimal algorithm correct?


igtrnt:
Is _opt_einsum.paths.optimal_ supposed to support non-standard Einsteins summations (when and index occurs in more than two tensors)?

Consider contracting 3 tensors: `ij,j,jk->`:

First, DFS looks at `ij,j` contraction. Resulting tensor is `j` (since `jk` is still in the remaining list).
So it caches `{ij,j} -> [j, cost1]`

Now, remaining contains: `jk` and `j` . Contracting them results in empty tensor.
So it caches `{j,jk} -> [, cost2]`.

Now DFS goes into another branch, starting with `j,jk` contraction. It should result in `j` tensor, because `ij` is still in the remaining list.
But cache contains `{j,jk} -> [, cost2]`  instead of `{j,jk} -> [j, *]`.

This simple example shows, that cache may not work correctly.
For larger examples, it's easy to end up with non-optimal "optimal" path and, moreover, completely wrong cost for that path.

jcmgray:
Hi @igtrnt yes I think you're right - the algorithm was written with 'standard' einsums in mind where two different intermediates sharing the exact same indices is not possible. Out of interest, do you have any more complex examples where this might be important? Inputs with single indices that can/should be immediately summed over (i and k here) are a known suboptimal edge case - #112, #114.

A workaround is just to use the `'dp'` path, which is optimal up to outer products (and scales better!). Or if you really want outer products you can instantiate `optimize=DynamicProgramming(search_outer=True)`. 

---

@dgasmith maybe this would be a good time to replace `optimal` with dynamic programming, either with `search_outer=True` or (preferable in my opinion) automatically off for larger contractions - sizes which optimal can't handle in any case. The only slight difference is that `'dp'` doesn't assume real FLOPs, but practically that doesn't matter... and maybe even makes slightly more sense.

igtrnt:
@jcmgray Here is a bigger example:
```
inputs = ["ABC", "D", "EF", "CE", "G", "H", "BH", "D", "BH"]
output = ''
sizes = {'A':9, 'B':11, 'C':7, 'D':24,'E':8, 'F': 2, 'G':5, 'H':30}
```

`optimal` ends up with the following contraction pass (SSA indexing):
```   
[(1, 7), (0, 9), (3, 10), (2, 11), (4, 12), (6, 8), (5, 13), (14, 15)]
```

[...]

Issue #114, opened 2019-11-30
Perform some simple preprocessing of contractions


jcmgray:
It might be sensible (#99 and #112) for ``opt_einsum`` to perform some simple universal pre-processing of contractions before sending them to the actual path optimizers. Including:

* **single index reduction** e.g. `abc->ab` where the index `c` appears on no other terms
* **hadamard de-duplication** e.g. `ab,ab->ab`

Some things to think about
- [ ] make the pre-processing optional (but default on?)
- [ ] might be useful to first have some generic way to combine bits of two paths
- [ ] include other more complex but sensible steps - e.g. contract disjoint subgraphs separately?

dgasmith:
This does need to be optional and I would think a default off state to retain current functionality. I am a bit hesitant to make changes to current algorithms without discussing more with downstream programs. Disjoint subgraphs seems like it would influence each algorithm differently and likely not able to be generalized. Similar to the path combination, each algorithm uses different internal structures at the moment and likely needs to remain that way for performance.

dgasmith:
@jcmgray I might try this today, are you working on any pieces currently?

jcmgray:
I'm adding a few tweaks to the 'dp' algorithm, (with an eye on it replacing the current 'optimal' implementation) but that's all pretty much self contained!

Issue #112, opened 2019-11-28
optimizing einsum('nlp,nlq->l', B, B)


yaroslavvb:
Is there a way to get optimized path for expression with some tensors repeated?

For instance, einsum('nlp,nlq->l', B, B) can be done in O(n^3) time, but einsum opt gives a schedule that takes O(n^4) time.
```
einsum_string = "nlp,nlq->l"
views = oe.helpers.build_views(einsum_string, {'n': 100, 'l': 100, 'p': 100, 'q': 100})
path, path_info = oe.contract_path(einsum_string, *views)
print(path_info)

Complete contraction:  nlp,nlq->l
         Naive scaling:  4
     Optimized scaling:  4
      Naive FLOP count:  2.000e+8
  Optimized FLOP count:  2.000e+8
   Theoretical speedup:  1.000
  Largest intermediate:  1.000e+2 elements
--------------------------------------------------------------------------------
scaling        BLAS                current                             remaining
--------------------------------------------------------------------------------
   4              0             nlq,nlp->l                                  l->l
```

Faster version way to do this
```
np.square(np.einsum('nlp->nl', B)).sum(dim=0)
```

jcmgray:
At the moment there is no way to do these non-linear types of contraction, only those that can be broken up into pairwise `tensordot`, `transpose` and `einsum` itself!

yaroslavvb:
Why doesn't opt_einsum discover the following schedule? It's 1000x faster ([colab](https://colab.research.google.com/drive/1zKh5v8bW3y-cWQlFX7HYVdgXLtPv2rQa#scrollTo=Iwdw21sFco_s))
```
x = np.einsum('nlp->nl', B)
np.einsum('ni,ni->i', x, x)
```

BTW, this einsum comes up when trying to extend https://arxiv.org/abs/1510.01799 to conv2d layers

dgasmith:
Currently `opt_einsum` only concerns itself with contract order and not the optimization of individual pair contractions. There is some notions that we could do this kind of optimization here, but the technology becomes less general quickly. Consider the following contraction `aij,ajk->aik` which for when `a` is large is best written as a looped GEMM. However on a CPU the threshold for the size of `a` is much smaller than for a GPU where the instantiation of different routines is higher unless batched GEMMs are called.

Happy to chat about adding this, but at the moment I do lean a bit towards downstream technologies implementing this kind of optimization themselves. I think we can do a decent job in NumPy, but less so with GPU backends.

[...]

PR #207, opened 2022-11-12
[WIP] add a preprocessor

jcmgray:
## Description

This adds a potential preprocessor that takes an equation and converts it into a minimal form with the following simplifications:

- All indices which only appear on a single input (and not the output) are summed over.
- All indices which appear multiple times on the same term are traced.
- All scalars are multiplied into the smallest other term
- All terms with the same indices are multiplied (hamadard product / elementwise) into a single term ('de-duplication').

along with a function, that transforms input arrays into the inputs for the new simplified `eq`. It is a function so that it could be part of a contract 'expression'.

This would address (#114, #99, #112, #167, #189, ...).

Example from the docstring:

```python
eq = ',ab,abee,,cd,cd,dd->ac'
arrays = helpers.build_views(eq)
new_eq, simplifier = make_simplifier(eq)
new_eq
# 'ab,cd,d->ac'
sarrays = simplifier(arrays)
oe.contract(new_eq, *sarrays)
# array([[0.65245661, 0.14684493, 0.42543411, 0.32350895],
#        [0.38357005, 0.08632807, 0.25010672, 0.19018636]])
```

## Todos
  - [ ] when and how should this be called?
  - [ ] return a partial path rather than a 'simplifier'?
  - [ ] should shape information (`size_dict`) be used and propagated for use with a contraction expression (and so that the smallest tensor to multiply scalars into can be most accurately chosen)?
  - [ ] other simplifications are possible (e.g. hadamard deduplication *up to transpose*, performing *all* non-rank increasing contractions)...
  - [x] ~currently the parsing is a single sweep, should it iteratively sweep til no more changes? E.g. currently examples like `ab,ab->` transform to `ab->` but if processed again could transform to `->`~

## Status
- [ ] Ready to go

### DIFF
--- a/opt_einsum/preprocessing.py
+++ b/opt_einsum/preprocessing.py
@@ -6,8 +6,7 @@ from .parser import find_output_str
 
 
 def multiply(*xs):
-    """Functional version of ``x[0] * x[1] * ...``.
-    """
+    """Functional version of ``x[0] * x[1] * ...``."""
     return functools.reduce(operator.mul, xs)
 
 
--- a/opt_einsum/preprocessing.py
+++ b/opt_einsum/preprocessing.py
@@ -1,0 +1,200 @@
+import operator
+import functools
+
+from .backends import get_func
+from .parser import find_output_str
+
+
+def multiply(*xs):
+    """Functional version of ``x[0] * x[1] * ...``.
+    """
+    return functools.reduce(operator.mul, xs)
+
+
+class Lazy:
+    """A tiny lazy computation class."""
+
+    __slots__ = ("inds", "f", "args")
+
+    def __init__(self, inds, f=None, *args):
+        self.inds = inds
+        self.f = f
+        self.args = args
+
+    def compute(self):
+        return self.f(
+            *(
+                arg.compute() if isinstance(arg, Lazy) else arg
+                for arg in self.args
+            )
+        )
+
+    def __repr__(self):
+        return f"Lazy('{self.inds}', {self.f}, {self.args})"
+
+
+def make_simplifier(*args, backend="numpy"):
+    """Take a einsum equation / specfification and return a new minimial
+    'processed' specfication and a function which will perform said processing.
+    The simplifications are as follows:
+
+        - All indices which only appear on a single input (and not the output)
+          are summed over.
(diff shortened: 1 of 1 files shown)


opt_einsum/preprocessing.py
```diff
@@ -0,0 +1,244 @@
+import operator
+import functools
+import collections
+
+from .backends import get_func
+from .parser import find_output_str
+
+
+def unique(it):
```
dgasmith, line 9: Can we type these? I cannot quite tell what this is doing, somewhat looks like a set operation.

opt_einsum/preprocessing.py
```diff
@@ -0,0 +1,244 @@
...
+
+    def __repr__(self):
+        return (
+            f"SimplifyExpression('"
+            f"{','.join(lz.inds for lz in self._lazy_inputs)}->"
+            f"{','.join(lz.inds for lz in self._lazy_outputs)}')"
+        )
+
+
+def make_simplifier(*args, backend="numpy"):
```
dgasmith, line 68:
```suggestion
def expression_simplifier(*args, backend="numpy"):
```

opt_einsum/preprocessing.py
```diff
@@ -0,0 +1,244 @@
...
+        >>> new_eq, simplifier = make_simplifier(eq)
+        >>> new_eq
+        'ab,cd,d->ac'
+
+        >>> sarrays = simplifier(arrays)
+        >>> oe.contract(new_eq, *sarrays)
+        array([[0.65245661, 0.14684493, 0.42543411, 0.32350895],
+               [0.38357005, 0.08632807, 0.25010672, 0.19018636]])
+    """
+    try:
```
dgasmith, line 112: We will need quite a few tests around this tech, too bad we cannot reproduce the original parsing tech here.

dgasmith (commented):
Makes sense, it would be good to start seeing some test here to find edge cases. A few notes:
- I would make a comment pass, see if we can highlight which operation we are reducing at each stage.
- The input/output pairing worries me a little bit. I wonder if we leave this as "internal" `oe.contract(..., preprocess=True)`
- It would be good to have a pathway for the `propocess=True` pathway.

Super cool addition!

PR #208, opened 2022-11-18
Use ConfigParser instead of SafeConfigParser

hongxu-jia:
The SafeConfigParser class will be renamed to ConfigParser in Python 3.12 [1]. This alias will be removed in future versions.So we can use ConfigParser directly instead.

[1] https://github.com/python/cpython/issues/89336

Signed-off-by: Hongxu Jia <hongxu.jia@windriver.com>

## Description
Provide a brief description of the PR's purpose here.

## Todos
  - [ ] TODO 1

## Questions
- [ ] Question1

## Status
- [ ] Ready to go


dgasmith (approved):
Looks good to me, fortunately this call is rendered out when a package was built and will not effect a package.

PR #222, opened 2023-11-24
Raise error in `parse_einsum_input` when output subscript is specified multiple times

lgeiger:
## Description
```python
contract("ij->jij", [[0, 0], [0, 0]])
# ValueError: einstein sum subscripts string includes output subscript 'j' multiple times
```
currently relies on the backend to throw an error if an output subscript is specified multiple times. However
```python
contract_path("ij->jij", [[0, 0], [0, 0]])
```
would still return an einsum path despite the wrong einsum equation. This might lead to subtle errors if a user ends up relying on this behaviour by accident. E.g. when using the jax backend no error is thrown but only an internal assertion fails.

This PR raises the error already in `parse_einsum_input` which ensures that `contract_path` matches the behaviour of `np.einsum` and all backends that might rely on `opt_einsum` for error handling will do so as well.

This PR also extends the unittests to test error handling for both `contract_path` and `contract` which is independent of the backend.

For reference I made a similar PR to https://github.com/numpy/numpy/pull/25230 to fix the same issue for `np.einsum_path`.

## Status
- [x] Ready to go

### DIFF
--- a/opt_einsum/parser.py
+++ b/opt_einsum/parser.py
@@ -379,8 +379,10 @@ def parse_einsum_input(operands: Any, shapes: bool = False) -> Tuple[str, str, L
     else:
         input_subscripts, output_subscript = subscripts, find_output_str(subscripts)
 
-    # Make sure output subscripts are in the input
+    # Make sure output subscripts are unique and in the input
     for char in output_subscript:
+        if output_subscript.count(char) != 1:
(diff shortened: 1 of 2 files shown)


dgasmith (approved):
Apologies for missing this, looks like a great change- especially in testing!

PR #232, opened 2024-05-20
Finalizes typing and updates documentation

dgasmith:
## Description
The PR provides a comprehensive typing overhaul which should provide a fully consistent typing paradigm. Given we will remove numpy as a dependency shortly (#204) we have several constraints on the typing update to avoid the `np.typing` module. While we can implement `ArrayType` as a `Protocol`, I'm unsure it's worth the effort.

## Todos
  - [ ] There are several overrides remaining which are required for MyPy to pass.

## Status
- [ ] Ready to go


jcmgray (commented):
All seems good to me, though I don't really use mypy/types in python so not an expert.

For the test failures, I guess the py3.9 ones are to do with https://peps.python.org/pep-0604/.

dgasmith:
@jcmgray Final tweaks are in, please re-review! I'll tackle #204 next.

Issue #235, opened 2024-06-13
`contract_path` with ellipsis fails when `shapes=True`


nova77:
Easy way to reproduce:

```python
input = np.random.rand(1,3,5,2,)
a = np.random.rand(2)
b = np.random.rand(7)

# ValueError: operands could not be broadcast together with shapes (4,) (0,) 
oe.contract_path('...D,D,k->...k', input.shape, a.shape, b.shape, shapes=True)
```
While if we do not pass the shapes, the function works:
```python
oe.contract_path('...D,D,k->...k', input, a, b)  # all good!
```
The source of error is in in [parser.py](https://github.com/dgasmith/opt_einsum/blob/master/opt_einsum/parser.py#L308) where `operands` (i.e. the shapes) are converted into arrays:
```python
operands = [possibly_convert_to_numpy(x) for x in operands[1:]]
```
The fix should be simple: just skip the conversion when `shapes` is True.

nova77:
The smallest change would be:
```python
    if shapes:
        operand_shapes = [list(s) for s in operands]  # from operand_shapes = operands
    else:
        operand_shapes = [o.shape for o in operands]
```

PR #236, opened 2024-06-14
Fixes `operands could not be broadcast` error when the subscripts contain ellipsis and the operands are shapes.

nova77:
## Description
When the subscript contain ellipsis and operands are shape, we get the following error:
```
ValueError: operands could not be broadcast together with shapes (..)
```
This is due to the fact that the parser converts the operands into numpy arrays which have strict rules on comparison.

This PR addresses https://github.com/dgasmith/opt_einsum/issues/235.

## Todos
  - [x] enforce operand shapes to be list of lists
  - [x] adds a unittest to verify the fixed behavior

## Status
- [ ] Ready to go

### DIFF
--- a/opt_einsum/parser.py
+++ b/opt_einsum/parser.py
@@ -309,8 +309,8 @@ def parse_einsum_input(operands: Any, shapes: bool = False) -> Tuple[str, str, L
     else:
         subscripts, operands = convert_interleaved_input(operands)
 
-    if shapes:
-        operand_shapes = operands
+   if shapes:
+        operand_shapes = [list(s) for s in operands]
     else:
         operand_shapes = [o.shape for o in operands]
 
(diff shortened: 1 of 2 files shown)


dgasmith:
This is a subtle bug related to a blocking issue in #204, where `possibly_convert_to_numpy`'s behavior is incorrect and we're not quite normalizing on internal representations the way we should. I'll add your test to #204 and see if I fix this in a more robust way for the next release.

dgasmith:
Yes, I can confirm #204 fixes it: https://github.com/dgasmith/opt_einsum/pull/204/commits/602e61ba38f43d4ab06501156ad6abd7ce002382

nova77:
Wonderful. I'm happy to have it fixed either way.

dgasmith:
Closed by #204.

PR #240, opened 2024-09-26
CI Fix

dgasmith:
## Description
Mamba-based CI appears to be broken. Trying a few things while keeping an eye on their issue tracker.

## Status
- [ ] Ready to go


loriab:
It's worth a try removing the two `mamba-version: "*"` lines since you get all the advantages of mamba from miniconda now.

dgasmith:
@loriab That works!

dgasmith:
Worked great!

PR #246, opened 2025-01-26
Added py.typed to signal this project uses type declarations

juanjosegarciaripoll:
## Description
When using opt_einsum in other projects, mypy complains that opt_einsum cannot be type checked. The reason is that for an installed library to be considered by mypy, a file py.typed must be installed in the root directory of the module.

## Status
- [x] Ready to go


dgasmith (commented):
Happy to merge this and push out a release. Before I do, can you manually insert a `py.typed` stub into an install and see how the typing goes?

I was slightly hesitant to add as I did not have a widely understood base of practice.

dgasmith:
@juanjosegarciaripoll gentle ping here.

PR #252, opened 2025-05-11
Fix: _einsum backend handling when optimize=False

DanisNone:
when calling opt_einsum.contract with optimize=False, opt_einsum._einsum did not receive any backend and used numpy by default. Now the logic has been corrected.


dgasmith:
Not sure why CI isn't triggering- can you push again as well as provide a test?

DanisNone:
@dgasmith Unfortunately, not sure what tests to add.

PR #257, opened 2026-03-14
Switches to UV

dgasmith:
## Description
Switches from Conda to UV for tests and CI. UV seems to resolve faster, be more canonical within the ML community, and has a number of nice bells.

## Status
- [ ] Ready to go


pyproject.toml
```diff
@@ -16,25 +16,44 @@ classifiers = [
...
+    "markdown==3.*",
+    "markdown-include==0.*",
+    "mkdocstrings[python]==0.29.*",
+    "mkdocs==1.*",
+    "mkdocs-awesome-pages-plugin==2.*",
+    "mkdocs-exclude==1.*",
+    "mkdocs-material==9.*",
+    "pygments==2.*",
+]
+dev = ["opt_einsum[test,lint]"]
```
j-wags: (not blocking) Is this sufficient for most dev envs, or should other deps be pulled in?

.github/workflows/Tests.yml
```diff
@@ -7,61 +7,50 @@ on:
...
 
-    - name: Environment Information
-      run: |
-        conda info
-        conda list
-        conda config --show-sources
-        conda config --show
+    - name: Install extra packages
+      if: matrix.extra-packages != ''
+      run: uv pip install ${{ matrix.extra-packages }}
```
j-wags: Installing `tensorflow-cpu` here is separate from the packages in the ephemeral env used in tests below. So I think tensorflow will need to be added to the `--with` list there.

pyproject.toml
j-wags: (not blocking) Is this sufficient for most dev envs, or should other deps be pulled in?

j-wags (approved):
LGTM, lots of little diffs from new relinting but the core changes (moving CI to UV) seems complete. Tensorflow tests can be re-enabled in another PR.

